PR #8, opened 2020-07-21
Tzirliya development

Tzirliya:
The basic structure of the Project Page is created. A simple css file is added. An index.html file is added so that project.html can be accessed from the dev server.


src/main/webapp/index.html
```diff
@@ -0,0 +1,17 @@
...
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Index</title>
+    <link rel="stylesheet" href="style.css">
+    <script src="script.js"></script>
+  </head>
+  <body>
+    <div id="navigation">
+      <a href="main_hub.html" id="hub">Main Hub</a>
```
talgsternberg, line 11: are index and main hub the same thing? or is the index page something else

Tzirliya: I guess it could be the same thing. I just created it so that when the dev server is run, there's a page that exists when it's first opened.

talgsternberg: Cool that works! This could potentially be a little description of what our website does/ an opportunity to login.

jcespejo: index.html, index.php, index.jsp, etc.. server fetches them automatically if you navigate into a directory without specifying an actual page.  ie. if you go "mysite.com/something" it would actually take you to "mysite.com/something/index.html" even if the page name is not explicitly specified

PR #10, opened 2020-07-21
Add buttons to Project Page

Tzirliya:
Buttons for admins, tasks, and members are added to the Project Page. CSS is used to make it formatted nicely. Feel free to change the color scheme though.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -2,13 +2,13 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>Main Hub</title>
+    <title>Login</title>
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
   <body>
     <div id="navigation">
-      <a href="index.html" id="hub">Main Hub</a>
+      <a href="main_hub.html" id="hub">Main Hub</a>
       <a href="user_profile.html" id="user">User Profile</a>
       <a href="project.html" id="project">Project Page</a>
       <a href="task.html" id="task">Task Page</a>
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -3,7 +3,7 @@
  */
 function goToUser() {
   // TODO: fetch for actual user's page
-  let url = 'user_profile.html';
+  const url = 'user_profile.html';
   location.href = url;
 }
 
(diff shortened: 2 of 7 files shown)


src/main/webapp/script.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Redirect to User Profile Page
+ */
+function goToUser() {
```
Tzirliya, line 4: This function and `goToUser()` are failing the checks because "they're defined but never used". However, they are actually being used in the project.html file.

src/main/webapp/index.html
```diff
@@ -2,13 +2,13 @@
...
     <meta charset="UTF-8">
-    <title>Index</title>
+    <title>Main Hub</title>
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
   <body>
     <div id="navigation">
-      <a href="main_hub.html" id="hub">Main Hub</a>
+      <a href="index.html" id="hub">Main Hub</a>
```
godsfavours, line 11: I was actually thinking the index.html page would be the page that directs the user to log in. I already made it, unless you feel strongly about making it tied to the main hub

src/main/webapp/script.js
talgsternberg: This might have absolutely nothing to do with it but I had an issue with assigning href to buttons vs. anchors and when I switched an element to an anchor it often worked. This looks completely correct but I'll do a deeper dive and see if I can find any potential issues.

src/main/webapp/index.html
Tzirliya: You're way's probably better. However, does the login page block any viewers from visiting the website on "guest mode", or should we allow people to view the website even if they're not logged in?

talgsternberg: I think if they're not logged in, their projects, tasks, checklist, etc are all empty

godsfavours: The way I imagine it, people would only have access to the website if logged in

jcespejo: > I was actually thinking the index.html page would be the page that directs the user to log in

If user was already logged in, would they be sent back to the login page again if they go back to the home (index) page?

There's also the question of users going directly to "site.com/task.jsp?12345".  Assuming the task ID is correct, how should this behave if they were logged in vs. if they were not logged in?

PR #21, opened 2020-07-23
New features on User_Profile.

talgsternberg:
Add a confirmation to edit user info (if everyone can see all user pages) and a skills section.

### DIFF
--- a/src/main/webapp/project.html
+++ b/src/main/webapp/project.html
@@ -19,10 +19,10 @@
       <h2>Admins</h2>
       <ul>
         <li class="inline">
-          <button type="button" onclick="goToUser()">Admin 1</button>
+          <button type="button" onclick="goToUser(1)">Admin 1</button>
         </li>
         <li class="inline">
-          <button type="button" onclick="goToUser()">Admin 2</button>
+          <button type="button" onclick="goToUser(2)">Admin 2</button>
         </li>
       </ul>
       <h2>Tasks</h2>
@@ -43,13 +43,13 @@
       <h2>Members</h2>
       <ul>
         <li class="inline">
-          <button type="button" onclick="goToUser()">User 1</button>
+          <button type="button" onclick="goToUser(1)">User 1</button>
         </li>
         <li class="inline">
-          <button type="button" onclick="goToUser()">User 2</button>
+          <button type="button" onclick="goToUser(2)">User 2</button>
         </li>
         <li class="inline">
-          <button type="button" onclick="goToUser()">User 3</button>
+          <button type="button" onclick="goToUser(3)">User 3</button>
         </li>
       </ul>
     </div>
--- a/src/main/webapp/project.html
+++ b/src/main/webapp/project.html
@@ -19,7 +19,6 @@
         <p>Default Project description...</p>
       </div>
       <h2>Admins</h2>
-<<<<<<< HEAD
       <ul>
         <li class="inline">
           <button type="button" onclick="goToUser(1)">Admin 1</button>
@@ -28,13 +27,10 @@
           <button type="button" onclick="goToUser(2)">Admin 2</button>
         </li>
       </ul>
-=======
       <div id="project-admins-container"></div>
->>>>>>> b4f09dae8dfcd261e6c6d77936c559e97132651d
       <h2>Tasks</h2>
       <div id="project-tasks-container"></div>
       <h2>Members</h2>
-<<<<<<< HEAD
       <ul>
         <li class="inline">
           <button type="button" onclick="goToUser(1)">User 1</button>
@@ -46,9 +42,7 @@
           <button type="button" onclick="goToUser(3)">User 3</button>
         </li>
       </ul>
-=======
       <div id="project-users-container"></div>
->>>>>>> b4f09dae8dfcd261e6c6d77936c559e97132651d
     </div>
   </body>
 </html>
--- a/src/main/webapp/project.html
+++ b/src/main/webapp/project.html
@@ -12,6 +12,7 @@
       <a onclick="goToUser()" id="user">User Profile</a>
       <a onclick="goToProject()" id="project">Project Page</a>
       <a onclick="goToTask()" id="task">Task Page</a>
+      <a onclick="goToSettings()" id="settings">Settings Page</a>
     </div>
     <div id="content">
       <div id="project-title-container"><h1>Default Project Name</h1></div>
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -184,8 +184,8 @@ function getUserInfo() {
       skillString = '';
       for (skill of users[user].skills) {
         if (skillString == '') {
-          skillString = skill.skill;}
-        else {
+          skillString = skill.skill;
+        } else {
           skillString = skillString + ', ' + skill.skill;}
       }
       skills.innerText = skillString;
@@ -196,11 +196,11 @@ function getUserInfo() {
 
 /**
  * Build ul element for subtasks on Task Page.
- * @param {Array} subtasks Array of taskIDs.
- * @return {Element} HTML ul element containing a list of subtasks.
  * Build return to project button.
  * @param {Hashmap} task Array of taskIDs.
  * @return {Element} HTML ul element containing a list of tasks.
+ * @param {Array} subtasks Array of taskIDs.
+ * @return {Element} HTML ul element containing a list of subtasks.
  */
 function getProjectReturn(task, subtasks) {
   const pElement = document.createElement('p');
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -171,26 +171,24 @@ function getUserInfo() {
       pskills = '';
       for (skill of users[user].skills) {
         if (skill.priority == true) {
-          if (pskills == '') {
-            pskills = skill.skill;
-          }
+          if (pskills == '') {pskills = skill.skill;}
           else {
-            pskills = pskills + ', ' + skill.skill
+            pskills = (pskills + ', ' + skill.skill);
           }
         }
       }
       prioritySkills.innerText = 'Priority Skills: ' + pskills;
       const skills = document.getElementById('user-skills-container');
-      skill_string = '';
-      for (skill of users[user].skills){
-        if(skill_string == ''){
-          skill_string = skill.skill;
+      skillString = '';
+      for (skill of users[user].skills) {
+        if (skillString == '') {
+          skillString = skill.skill;
         }
-        else{
-          skill_string = skill_string + ', ' + skill.skill
+        else {
+          skillString = skillString + ', ' + skill.skill
         }
       }
-      skills.innerText = skill_string;
+      skills.innerText = skillString;
       break;
     }
   }
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -173,8 +173,7 @@ function getUserInfo() {
         if (skill.priority == true) {
           if (pskills == '') {
             pskills = skill.skill;
-          }
-          else {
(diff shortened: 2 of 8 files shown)


src/main/webapp/user_profile.html
```diff
@@ -14,7 +14,15 @@
...
-      <h1>Welcome, UserName</h1>
+      <h1>UserName</h1>
+      <div>Are you UserName? Re-confirm to make edits to your page</div>
+      <form>
+          <label for="email">email</label><br>
+          <input type="text" id="email" name="email" value="test@gmail.com"><br>
+          <label for="password">password:</label><br>
+          <input type="text" id="password" name="password" value="test"><br><br>
+          <input type="submit" value="Submit">
+      </form>
```
Tzirliya, line 25: In the future, should this be a login button that redirects to a login page (using the authorization API)?

talgsternberg: I was thinking the person is already logged in (or else they would have no way to access their page) but this is just an extra measure of security (like when you have to re-enter password on a Mac to change some settings even though you are already logged in). That was just my thinking but I'm open to change!

jcespejo: imo it would depend on what they're changing.  If they're changing passwords (or maybe email as well), I would require them to reenter their password (just the password though, the session should already know about the username/email).  If they're changing something else, it might be too much of a hassle to login again.

src/main/webapp/user_profile.html
```diff
@@ -14,16 +14,34 @@
...
         <div>Year: Junior</div>
         <div>Major: Biology</div>
         <div>Total Completed Tasks: 7</div>
         <button type="button" onclick="goToTask()">See Tasks</button>
       </div>
-      <h2>Checklist:</h2>
+      <h2>User Skills:</h2>
+      <div class="user-stats">
+        <div>Leadership, speaks 3 languages, HTML/CSS, organization</div>
+        <div> </div>
```
Tzirliya, line 37: What's the purpose of this `<div>`?

src/main/webapp/user_profile.html
```diff
@@ -14,16 +14,34 @@
...
+        <form id="form">
+          <label class="sub-form" for="skills">Update Your Skills</label><br>
+          <input type="text" id="skills" name="skills" value="ex: leadership, Python, statistics..."><br>
+          <input class="sub-form" type="submit" value="Submit">
+        </form>
```
Tzirliya, line 42: Maybe be more specific with the form id. Maybe `<form id="skills-form>`?

src/main/webapp/user_profile.html
talgsternberg: > This page seems to be doing both "User Settings" and "User Profile/Activity". The difference is the former is private and should only be visible for the user. The later is public and could be visible for everyone. We also don't want them to accidentally change settings while viewing their active tasks here.

Would you suggest having 2 pages instead? or just separating them?

jcespejo: separating into 2 pages.

talgsternberg:
Any suggestions for how to make a skill a "top" skill? I'm not sure the best/most visually appealing way to do this...

src/main/webapp/script.js
```diff
@@ -31,11 +33,19 @@ const gJSONtasks = JSON.stringify(gTasks);
...
+    major: 'Studio Art', numTaskCompleted: 5, admin: true},
+  user2: {userID: 2, name: 'User 2', skills: 'Object Oriented Programming',
+    major: 'Computer Science', numTaskCompleted: 8, admin: false},
+  user3: {userID: 3, name: 'User 1', skills: 'Leadership, Organization',
+    major: 'Chemistry', numTaskCompleted: 4, admin: false}};
+
+const gJSONusers = JSON.stringify(gUsers);
+const gDefaultUser = {userID: 0, name: 'Default Username',
+  skills: 'Default Skills', major: 'Default Major',
+  numTaskCompleted: 'Default Number', admin: false}; // Add attributes
```
Tzirliya, line 48: I would put multiple skills as a list of strings, like `skills: ['Art', 'Writing']`. Same with major if double majoring will be considered.

Tzirliya, line 48: What does admin refer to here? Users can be part of multiple tasks and projects, and they won't necessarily be admin for all or none. I'm storing a list of admins with projects instead.

src/main/webapp/user_profile.html
```diff
@@ -6,40 +6,73 @@
...
-      <div id="user-stats">
-        <div>Name: UserName LastName</div>
-        <div>Year: Junior</div>
-        <div>Major: Biology</div>
-        <div>Total Completed Tasks: 7</div>
-        <button type="button" onclick="goToTask()">See Tasks</button>
+      <div class="user-stats">
+        <div id="year-container">Year: Default Year</div>
+        <div id="major-container">Major: Default Major</div>
+        <div id="num-complete-container">Total Completed Tasks: default</div>
```
Tzirliya, line 22: I styled the container ids as "page-type-container", for example "project-admins-container". You don't have to do the same thing, but you can change the ids to something like "user-year-container" for consistency. (And this helps avoiding confusion just in case someone would want to make a "year-container" on another page.)

godsfavours:
> Any suggestions for how to make a skill a "top" skill? I'm not sure the best/most visually appealing way to do this...

In the JS file, you could make the skills a list like Tzirliya suggested in one of the comments. Instead of just a string you could make it an array of skill objects like so: [{name = 'Art', priority = false}, {name = 'Writing', priority = true}]. Then to display it you could just add an asterisks next to the name if priority is set to true

src/main/webapp/user_profile.html
```diff
@@ -6,40 +6,73 @@
...
+          <label for="art">Artistic Ability</label><br>
+          <input type="checkbox" id="webdev" name="webdev" value="webdev">
+          <label for="webdev">Web Development (HTML,CSS,Javascript)</label><br>
+          <input type="checkbox" id="oop" name="oop" value="oop">
+          <label for="oop">Object Oriented Programming Languages (ex:java)</label><br><br>
+          <input type="submit" value="Submit">
+        </form>
+        <h3>Change Other Information:</h3>
+        <form id="stats-form">
+          <input type="text" id="name" name="name" value="ex: John Doe"><br>
```
jcespejo, line 49: supernit:  I don't think we need to provide an example of a name.

src/main/webapp/user_profile.html
```diff
@@ -6,40 +6,73 @@
...
+          <label for="name">Update Name:</label><br>
+          <input type="text" id="year" name="year" value="ex: Junior"><br>
```
jcespejo, line 51: I thought we were going to be doing "2023" instead of "Junior" etc..

src/main/webapp/user_profile.html
```diff
@@ -6,40 +6,73 @@
...
+          <input type="checkbox" id="organization" name="organization" value="organization">
+          <label for="organization">Organization</label><br>
+          <input type="checkbox" id="writ" name="writ" value="writ">
+          <label for="writ">Analytical/Creative Writing</label><br>
+          <input type="checkbox" id="art" name="art" value="art">
+          <label for="art">Artistic Ability</label><br>
+          <input type="checkbox" id="webdev" name="webdev" value="webdev">
+          <label for="webdev">Web Development (HTML,CSS,Javascript)</label><br>
+          <input type="checkbox" id="oop" name="oop" value="oop">
+          <label for="oop">Object Oriented Programming Languages (ex:java)</label><br><br>
```
jcespejo, line 44: ok for the prototype, I would suggest not to use br for formatting your page though, since semantically br would be more for breaking a line of text into multiple lines.  Maybe ul/li would be more appropriate.

src/main/webapp/project.html
```diff
@@ -27,9 +28,13 @@ <h2>Admins</h2>
           <button type="button" onclick="goToUser(2)">Admin 2</button>
         </li>
       </ul>
+=======
+      <div id="project-admins-container"></div>
+>>>>>>> b4f09dae8dfcd261e6c6d77936c559e97132651d
```
talgsternberg, line 33: not sure whats going on here....

src/main/webapp/script.js
```diff
@@ -93,6 +95,21 @@ function getUserInfo() {
...
+          if(pskills == ''){
+            pskills = skill.skill;
+          }
+          else{
+            pskills = pskills + ', ' + skill.skill
+          }
+        }
+      }
+      console.log('pskills:');
+      console.log(pskills);
```
talgsternberg, line 187: pskills outputs in console what I want it to, yet the container shows [object,Object],[object,Object]. Any ideas why? I can't seem to figure this out...

src/main/webapp/script.js
```diff
@@ -95,6 +95,32 @@ const gJSONusers = JSON.stringify(gUsers);
...
+ * the URL, default values will be shown.
+ */
+function getProjectInfo() {
+  const params = new URLSearchParams(location.search);
+  const projectID = params.get('projectID');
+  const projects = JSON.parse(gJSONprojects);
+  for (project in projects) {
+    if (projects[project].projectID == projectID) {
+      const title = document.getElementById('project-title-container');
+      title.innerHTML = '<h1>' + projects[project].name + '</h1>';
```
talgsternberg: I didn't add this but when I pulled from master branch this line had an error and my whole javascript file was a bit messed up. Not sure what happened. Did I pull at the wrong time?

src/main/webapp/script.js
```diff
@@ -6,12 +6,23 @@ function goToHub() {
...
+}
+
 /**
  * Redirect to User Profile Page.
+ * if user is going to their own page (default)
+ * then eventually we pass their UserID
+ * @param {number} userID
  */
-function goToUser() {
+function goToUser(userID) {
```
Tzirliya, line 23: Perhaps change it to an optional parameter (temporarily)? Because sometimes goToUser() is called without a parameter.

jcespejo:
This is turning into a pretty big PR.  I would suggest finding a good stopping point and then merging this into master before continuing on

src/main/webapp/script.js
```diff
@@ -133,11 +155,63 @@ function getTaskInfo() {
...
+/**
+ * Build ul element for subtasks on Task Page.
  * Build return to project button.
  * @param {Hashmap} task Array of taskIDs.
  * @return {Element} HTML ul element containing a list of tasks.
+ * @param {Array} subtasks Array of taskIDs.
+ * @return {Element} HTML ul element containing a list of subtasks.
  */
-function getProjectReturn(task) {
+function getProjectReturn(task, subtasks) {
```
Tzirliya, line 214: Why is subtasks added?

src/main/webapp/script.js
```diff
@@ -158,6 +232,7 @@ function getProjectReturn(task) {
  * Build ul element for tasks.
  * @param {Array} tasks Array of taskIDs.
  * @return {Element} HTML ul element containing a list of tasks.
+>>>>>>> b4f09dae8dfcd261e6c6d77936c559e97132651d
```
Tzirliya, line 235: Is this left over from the merge?

talgsternberg:
> This is turning into a pretty big PR. I would suggest finding a good stopping point and then merging this into master before continuing on

Okay will do! I just need one more approval.

src/main/webapp/script.js
talgsternberg: Yep! I missed that one, oops!

src/main/webapp/script.js
talgsternberg: For some reason javascript lint was getting mad when it wasn't because it was defined as a param ... no worries if you change it back I just wanted to pass lint tests for this PR   :)

PR #22, opened 2020-07-23
Tzirliya development

Tzirliya:
The Task Page shows details for a specific task (one of three hard-coded tasks or a default task if accessed via the navigation bar). These details now include a list of subtasks (buttons and descriptions) and a list of users (buttons).


src/main/webapp/script.js
```diff
@@ -42,7 +53,95 @@ function getTaskInfo() {
...
+      const subtasks = document.getElementById('task-subtasks-container');
+      subtasks.appendChild(getSubtasks(tasks[task].subtasks));
+      const users = document.getElementById('task-users-container');
+      users.appendChild(getUsers(tasks[task].users));
       break;
     }
   }
+}
+
+/**
```
jcespejo, line 65: Good looking documentation.  JSDoc?  Let's definitely keep this up.

PR #25, opened 2020-07-24
Create TaskData and TaskController objects

Tzirliya:
TaskData.java is created. It holds all details about a task that would be stored in Datastore. TaskController.java is a class that limits the actions performed on TaskData. The Java servlets will only interact with the TaskController object.


talgsternberg:
Hi! Can you describe in slightly more detail what TaskController does? I think I understand I just want to make sure I'm on the same page.

Tzirliya:
The TaskData object holds all task attributes and all getters and setters. The TaskController object has a TaskData object. TaskController limits the actions that can be performed on TaskData by choosing which getters and setters should be available and by creating more limited methods like addSubtask() and removeSubtask() instead of allowing the servlet code to change the subtasks all at once. The Java servlets will only interact with the TaskController and its limited methods iiuc.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,82 @@
...
+  public void removeUser(long userID) {
+    HashSet<Long> users = taskData.getUsers();
+    users.remove(userID);
+    taskData.setUsers(users);
+  }
+
+  public void addSubtask(long taskID) {
+    HashSet<Long> subtasks = taskData.getSubtasks();
+    subtasks.add(taskID);
+    taskData.setUsers(subtasks);
```
godsfavours, line 73: I think you mean to call setSubtasks(subtasks) here

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,82 @@
+package com.rtb.projectmanagementtool.task;
+
+import java.util.HashSet;
+
+/** Class controlling the TaskData object. */
+public final class TaskController {
+
+  private TaskData taskData;
```
jcespejo, line 8: Sorry to clarify, TaskController doesn't necessarily have to map to only a single TaskData.  Your TaskController can (and normally would) have a list/map of TaskDatas.

Think of the controller as the thing that does the operations on one or more tasks, specifically for example load/save/add/delete tasks.  So most probably this will also hold all of your Datastore load/save logic.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,82 @@
...
+  public void addSubtask(long taskID) {
+    HashSet<Long> subtasks = taskData.getSubtasks();
+    subtasks.add(taskID);
+    taskData.setUsers(subtasks);
+  }
+
+  public void removeSubtask(long taskID) {
+    HashSet<Long> subtasks = taskData.getSubtasks();
+    subtasks.remove(taskID);
+    taskData.setUsers(subtasks);
```
godsfavours, line 79: and here

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,82 @@
...
+/** Class controlling the TaskData object. */
+public final class TaskController {
+
+  private TaskData taskData;
+
+  public TaskController(TaskData taskData) {
+    this.taskData = taskData;
+  }
+
+  // Is there a way to avoid retyping all of these getters and setters?
```
jcespejo, line 14: Eclipse has a "generate all getters/setters" functionality.  I'm not sure if the Cloud Editor has something similar (probably not).  But as per comment above, you won't need to duplicate here since field getters/setters would only be needed on the Data object.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,82 @@
+package com.rtb.projectmanagementtool.task;
```
jcespejo, line 1: Yup

src/main/java/com/rtb/projectmanagementtool/task/TaskData.java
```diff
@@ -0,0 +1,94 @@
...
+    this.users = users;
+  }
+
+  public void setSubtasks(HashSet<Long> subtasks) {
+    this.subtasks = subtasks;
+  }
+
+}
+
+enum Status {
```
jcespejo, line 91: I would suggest moving these up higher, maybe even before the class, just so they're visible right away.  Another option would be to move these into a "constants" file maybe in com.rtb.projectmanagementtool.util.Constants.java

src/main/java/com/rtb/projectmanagementtool/task/TaskData.java
```diff
@@ -0,0 +1,88 @@
...
+  private long taskID;
+  private long projectID;
+  private String name;
+  private String description;
+  private Status status;
+  private HashSet<Long> users;
+  private HashSet<Long> subtasks;
+
+  public TaskData(long taskID, long projectID, String name, String description,
+    Status status, HashSet<Long> users, HashSet<Long> subtasks) {
```
godsfavours: maybe add different constructor(s) for if some of these arguments aren't sent in

Tzirliya:
Is this closer to what TaskController is supposed to look like?

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,80 @@
...
+
+  // Optional TODO: Add sort and quantity paramters.
+  public HashSet<TaskData> getAllTasks() {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    HashSet<TaskData> tasks = new HashSet<>();
+    Query query = new Query("Task");
+    PreparedQuery results = datastore.prepare(query);
+    for (Entity entity : results.asIterable()) {
+      // Get task entity data
+      long taskID = (long) entity.getKey().getId();
```
jcespejo, line 28: I would move these to the Data layer, maybe pass an Entity as a param to its constructor.  It'll be like the Data object building itself based on a database entity.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,80 @@
...
+      HashSet<Long> users = task.getUsers();
+      HashSet<Long> subtasks = task.getSubtasks();
+      // Build task entity
+      Entity entity = new Entity("Task", taskID);
+      entity.setProperty("projectID", projectID);
+      entity.setProperty("name", name);
+      entity.setProperty("description", decription);
+      entity.setProperty("status", status);
+      entity.setProperty("users", users);
+      entity.setProperty("subtasks", subtasks);
```
jcespejo, line 61: Also these would be good on the Data layer, so the Data layer can transform itself to and from an entity.

jcespejo:
> Is this closer to what TaskController is supposed to look like?

Yep this looks good, with the TaskController doing the add/edit/delete operations.  Other things that could fit in there is association methods (add a user to a task), search/sort/filter, etc..

Only change I would do is to move the `Data <-> Entity` translation in the Data itself (either via a constructor param, or a `ParseFromEntity()` and `ToEntity()` methods).  If it helps to clarify, the Entity is *the* data, and TaskData is a wrapper around the entity to make it easier to work with.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public TaskController(HashSet<TaskData> tasks) {
+    this.tasks = tasks;
+  }
+
+  // Optional TODO: Add sort and quantity paramters.
+  public HashSet<TaskData> getAllTasks() {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    HashSet<TaskData> tasks = new HashSet<>();
+    Query query = new Query("Task");
```
jcespejo, line 24: ok for the first pass.  But also think about if we really do need to load all tasks every time.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -0,0 +1,80 @@
...
+      entity.setProperty("subtasks", subtasks);
+      taskEntities.add(entity);
+    }
+    datastore.put(taskEntities);
+  }
+
+  public void deleteTasks(HashSet<Long> taskIDs) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    HashSet<Key> keys = new HashSet<>();
+    Query query = new Query("Comment").setKeysOnly();
```
jcespejo, line 70: Same here, if we truly need to load all tasks to delete one.

[Looking at the Query API](https://googleapis.dev/java/google-cloud-datastore/latest/com/google/cloud/datastore/Query.html), I see there's a GqlQuery or a KeyQuery options.  GqlQuery I'm assuming would allow you to build SQL formatted queries?  Then maybe KeyQuery allows an entity key to query for?

PR #26, opened 2020-07-27
new user class

talgsternberg:
Do I need a UserController type of class too where I should define the non getter/setter methods?

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/user/UserData.java
+++ b/src/main/java/com/rtb/projectmanagementtool/user/UserData.java
@@ -3,7 +3,7 @@ package com.rtb.projectmanagementtool.user;
 import java.util.HashSet;
 
 /** Enum containing skills for user. */
-public enum Skills {
+enum Skills {
   LEADERSHIP (false),
   ORGANIZATION (false),
   WRITING (false),
--- a/src/main/java/com/rtb/projectmanagementtool/user/UserData.java
+++ b/src/main/java/com/rtb/projectmanagementtool/user/UserData.java
@@ -4,12 +4,12 @@ import java.util.HashSet;
 
 /** Enum containing skills for user. */
 enum Skills {
-  LEADERSHIP (false);
-  ORGANIZATION (false);
-  ANALYTICAL/CREATIVE WRITING (false);
-  ARTISTIC ABILITY (false);
-  WEB DEVELOPMENT (false);
-  OBJECT ORIENTED PROGRAMMING (false);
+  LEADERSHIP (false),
+  ORGANIZATION (false),
+  ANALYTICAL/CREATIVE WRITING (false),
+  ARTISTIC ABILITY (false),
+  WEB DEVELOPMENT (false),
+  OBJECT ORIENTED PROGRAMMING (false)
 }
 
 /** Class for User data. */
--- a/src/main/java/com/rtb/projectmanagementtool/user/UserData.java
+++ b/src/main/java/com/rtb/projectmanagementtool/user/UserData.java
@@ -4,7 +4,12 @@ import java.util.HashSet;
 
 /** Enum containing skills for user. */
 enum Skills {
-  //not sure what to add here
+  LEADERSHIP (false);
+  ORGANIZATION (false);
+  ANALYTICAL/CREATIVE WRITING (false);
+  ARTISTIC ABILITY (false);
+  WEB DEVELOPMENT (false);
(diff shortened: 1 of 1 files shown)


src/main/java/com/rtb/projectmanagementtool/user/UserData.java
```diff
@@ -0,0 +1,79 @@
+package com.rtb.projectmanagementtool.user;
+
+import java.util.HashSet;
+
+/** Enum containing skills for user. */
+enum Skills {
+  //not sure what to add here
```
talgsternberg, line 7: Do I leave it blank or create an empty HashMap<String>? I've never used enum and am a bit unclear.

jcespejo: enums doc:  https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

Specifically look at how they build the enum "class" for planets.  that'll allow you to specify a string description for your enum values.

src/main/java/com/rtb/projectmanagementtool/user/UserData.java
```diff
@@ -4,12 +4,12 @@
...
-  ARTISTIC ABILITY (false);
-  WEB DEVELOPMENT (false);
-  OBJECT ORIENTED PROGRAMMING (false);
+  LEADERSHIP (false),
+  ORGANIZATION (false),
+  ANALYTICAL/CREATIVE WRITING (false),
+  ARTISTIC ABILITY (false),
+  WEB DEVELOPMENT (false),
+  OBJECT ORIENTED PROGRAMMING (false)
 }
```
talgsternberg, line 14: based on lint errors, it's almost as if it expects another class/enum in between my enum and the comment. Not sure what's wrong with my formatting here...

jcespejo: Couple things, enums can also be used as named variables, so you'll want to keep consistent with variable naming conventions (no spaces, special chars, etc..).

Then, I'm curious what's the intention of the bool here?  What are we tracking with those flags?

Finally, please check the [public enum Planet](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) example again.  They're passing in 2 numbers into the enum, the rest of the class could provide clues as to what those two numbers are.

talgsternberg: the booleans represent the "priority" or "specialized" skills.

jcespejo: Good thought.  In that case the bool is ok, you'll just need to make sure it's included in the constructor.

Also tbh I'm not familiar with how enums are represented in the Datastore.  Another option could be to build these as actual Data Entities and associate them to the user via the SkillsId

godsfavours: I think the best way to go about this would be to omit the boolean flag and just have the enum as:
enum Skill {
    WEB_DEVELOPMENT,
    ARTISTIC_ABILITY,
    etc.
}
then you can have different arrays to store the different categories of skills:
HashSet<Skill> prioritySkills;
HashSet<Skill> skills;

If you wanted to keep the enums inside one array and include the boolean flag in the enum (I'm not sure if that is good practice),
you'd have to make the enum like so:

enum Skill {
    WEB_DEVELOPMENT(false),
    ARTISTIC_ABILITY(true),
    etc.

     private final boolean isPriority;

    private Skill(boolean isPriority) {
        this.isPriority = isPriority;
    }
}

jcespejo:
> Do I need a UserController type of class too where I should define the non getter/setter methods?

It might be helpful to include in the design doc a list of operations you'd need to do with Users.

src/main/java/com/rtb/projectmanagementtool/user/UserData.java
jcespejo: @godsfavours:  if you have the two lists though, that'll mean flagging a skill as priority would be two operations:  delete from the non-priority list then add into the priority list.  Depending on how these are written to the database, some engines might write them separately, and if the operation failed midway, that'll mean you'll have inconsistent state (eg. having the same skill in both lists).


Alternative:  What will we be using skills for for the MVP.  Maybe we can instead leave it out and add later on once the core relationships are built.

talgsternberg: Okay. Should I keep skills for now but just move on from "priority" until we finish MVP?

PR #30, opened 2020-07-27
Project & ProjectController classes

godsfavours:
Same pull request from earlier. Made a new one to clean up and better organize commits


src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import java.util.HashSet;
+
+public class ProjectController {
+  private DatastoreService datastore;
```
jcespejo, line 17: Probably would be better to pass this from outside, rather than having a local instance here.

For most database engines, having a "database client" or "database connection" would mean it'll hold on to the connection until the process is done.  Typically you'd want to limit the number of connections you have open to save on resources, so it's usually advisable to have 1 connection and pass it around (via params to db consumers), rather than having one connection per consumer.

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,101 @@
...
+  private HashSet<ProjectData> projects;
```
jcespejo, line 18: This may be ok.  But also think about what you'd do with this list of projects later on.  If you'll be passing this around to other consumers (controllers or servlets), it might be best to not own an instance here and instead just pass the container in when doing load/create/remove operations.  This way you wouldn't have to worry about the data being copied when passing between consumers.

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,101 @@
...
+  /** Class constructor */
+  public ProjectController() {
+    datastore = DatastoreServiceFactory.getDatastoreService();
+    projects = new HashSet<ProjectData>();
+    loadProjects();
+  }
+
+  /** Initializes the projects HashSet by retrieving all datastore entities of type "Project" */
+  private void loadProjects() {
+    PreparedQuery results = datastore.prepare(new Query("Project"));
```
jcespejo, line 29: Do we always have to load all the projects every time?  Maybe also think about cases where you'd only need a subset.

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Creates a new project First creates the Project entity and then creates the project object with
+   * the projectId returned by createProjectEntity()
+   *
+   * @param name the name of the project
+   * @param description the description of the project
+   */
+  public void createProject(String name, String description, long creatorId) {
+    ProjectData project = new ProjectData(name, description, creatorId);
+    projects.add(project);
```
jcespejo, line 46: This wouldn't save the new project into the datastore though, correct?  Will that be added later on?

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Retrieves the Project object with the matching projectId from HashSet, or nothing if the
+   * project is not found
+   *
+   * @param projectId the id of the project to retrieve
+   * @return the project if found, or null if not found
+   */
+  public ProjectData getProject(long projectId) {
```
jcespejo, line 89: How often do you see this method being used?  If a lot, is there another way  we can store the projects to make lookups faster?

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * @param projectId the id of the project to remove
+   * @return true if the removal is successful
+   */
+  private boolean removeProjectEntity(long projectId) {
+    try {
+      Key projectEntityKey = KeyFactory.createKey("Project", projectId);
+      datastore.delete(projectEntityKey);
+      return true;
+    } catch (Exception exception) {
```
jcespejo, line 75: Good use of try/catch.  This is definitely something I'd advice adding into all datastore operations.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+   * constant variables used to specify the names of each property of the Project entity. Used to
+   * minimize redundancy and also make renaming properties easy if needed
+   */
+  private final String PROPERTY_NAME = "name";
+
+  private final String PROPERTY_DESCRIPTION = "description";
+  private final String PROPERTY_USER_IDS = "userIds";
+  private final String PROPERTY_TASK_IDS = "taskIds";
+
+  private DatastoreService datastore;
```
jcespejo, line 35: See note I had above regarding limiting the number of "connections" you have open.  Another thing to think about, you'll have multiples of this ProjectData items and the DatastoreService object most probably is pretty heavy.  We wouldn't want to use up resources to have multiples of these

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Class constructor with the minimum requirements for creating a project.
+   *
+   * @param name the name of the project
+   * @param description the description of the project
+   * @param creatorId the id of the user who created the project
+   */
+  public ProjectData(String name, String description, long creatorId) {
+    datastore = DatastoreServiceFactory.getDatastoreService();
```
jcespejo, line 51: I would advise separating the database operations from your Data model object.  One way to reason about this is by thinking about the "separation of responsibilities".  You'd want your Data model to be just that, a model/representation for the data.  You'd then want operations done to copes of these done somewhere else (the controller for instance).

Another thought, separating them would allow us later on if we decide to swap to a different flavor of Datastore (or maybe even a different database engine all together), you won't have to update all of your Data models.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+    // sets the type of containers to store parsed JSON in
+    Type mapType = new TypeToken<HashMap<Long, User>>() {}.getType();
+    Type setType = new TypeToken<HashSet<Long>>() {}.getType();
+
+    this.userIds = gson.fromJson((String) entity.getProperty(PROPERTY_USER_IDS), mapType);
+    this.taskIds = gson.fromJson((String) entity.getProperty(PROPERTY_TASK_IDS), setType);
+  }
+
+  /** Saves the object as a datastore entity. */
+  private void toEntity() {
```
jcespejo, line 86: `toSomething` naming conventions usually implies conversion and returning the converted value.  think about `toString` that would return a string conversion of an object, etc..

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+    Gson gson = new Gson();
+    try {
+      Key entityKey = KeyFactory.createKey("Project", this.id);
+      Entity entity = datastore.get(entityKey);
+
+      entity.setProperty(PROPERTY_NAME, this.name);
+      entity.setProperty(PROPERTY_DESCRIPTION, this.description);
+      entity.setProperty(PROPERTY_USER_IDS, gson.toJson(this.userIds));
```
jcespejo, line 94: This is an interesting idea.  While storing as json would allow you to store and later on read more formatted data, this will also require parsing later on.  This might work out well (maybe) if we're storing child objects that has multiple fields.  In this specific scenario though, since we're only storing a list of IDs, try to see if there's a better alternative.

Also keep in mind one caveat with storing as json is you wouldn't be able to query against this field.  For example if project userIDs is stored as json, it'll be hard to query for "all projects user X owns"

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+  @Override
+  public String toString() {
+    String returnString = "{\n";
+    returnString += "Project id: " + this.id + "\n";
+    returnString += "Project Name: " + this.name + "\n";
+    returnString += "Project Description: " + this.description + "\n";
+    returnString += "Project Users: " + this.userIds.toString() + "\n";
+    returnString += "Tasks: " + this.taskIds.toString() + "\n}";
+    return returnString;
+  }
```
jcespejo, line 203: Definitely useful to provide a toString method

src/main/java/com/rtb/projectmanagementtool/project/ProjectTestServlet.java
```diff
@@ -0,0 +1,37 @@
+// Just using this Servlet to test the Project & ProjectController classes.
+// I didn't know any other ways to test the classes other than through unit tests, which I
+// couldn't get working.
+// Feel free to mess around with it and test some of the functionality/see how it can be improved
+
+package com.google.sps.servlets;
```
jcespejo: Please watch out for package name vs. directory structure.  In most cases they would have to be the same.

src/main/java/com/rtb/projectmanagementtool/project/ProjectTestServlet.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/my-project-test")
+public class ProjectTestServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
jcespejo: Good idea putting together a test for these.  I would suggest starting with unit tests first though before putting together a servlet.

Also keep in mind servlets are accessible to the public (unless if you setup access restrictions), so you wouldn't want to have test logic exposed publicly.

Finally It would be best to run tests off of a test Datastore instance.  This way you won't create test data into your prod database.

src/main/webapp/script.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * Redirect to Main Hub Page.
+ */
+function projectTest() {
+  fetch('my-project-test');
```
jcespejo: You can also just navigate to `site.com/my-project-test` to see your servlet's output without having to hook it up to your pages.

src/main/webapp/index.html
```diff
@@ -6,7 +6,7 @@
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
-  <body>
+  <body onload="projectTest()">
```
jcespejo: Watch out for this since this'll add test code to shared resources.  ie. your teammates would end up pulling this test into their dev environments.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.HashSet;
+
+enum User {
+  CREATOR,
+  ADMIN,
```
emmacollins, line 20: what's the difference between a creator and admin? might be useful to add a few comments here explaining each type of User.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,204 @@
...
+   *
+   * @param description the new description of the project
+   */
+  public void setDescription(String description) {
+    this.description = description;
+    toEntity();
+  }
+
+  /**
+   * Adds a user to the project ? separate this into addAdminUser and addRegularUser?
```
emmacollins, line 149: I'm in favor of separating into two methods. Passing booleans around quickly gets tedious and sometime dangerous (i.e. if a caller accidentally passes two boolean values into a method out of order it can lead to unintended behavior).

godsfavours:
> Good start. I would suggest at least removing the test servlet from your "prod" flow.

test servlet removed

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
godsfavours: I can't think of a use case for this method rn, but I think it might be good to have just in case. I've changed the method to retrieve the project directly from datastore

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
godsfavours: Could this method (or some other form of it) be used to not just create new entitites but also to update existing ones? 
The problem with returning an entity here is that when we try to add it to the datastore in ProjectController, a whole new entity will be created, and that isn't ideal if we just want to update the entity's properties.

jcespejo: > The problem with returning an entity here is that when we try to add it to the datastore in ProjectController, a whole new entity will be created, and that isn't ideal if we just want to update the entity's properties.

I don't think we necessarily have to push Entities directly into the Datastore as soon as they're created.  Entities, being objects themselves, can be passed around, stored, or operated on later on.  This will allow us to better separate your Data Model from your operations.

I think the question you're thinking about here illustrates a good use case for why you'd want to separate the Model from your Controller.  The entity itself (ProjectData), shouldn't really care if it's being added or being updated.  All it cares about is how its fields are filled in, and how it shares it fields to other consumers.

What you do with that Data is someone else's (Controller's) responsibilities.  That someone else is the one that needs to keep track of the difference between an insert and an update.

godsfavours: Okay that makes sense

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -66,38 +50,43 @@ public ProjectData(String name, String description, long creatorId) {
...
+    Entity entity = new Entity("Project");
+    entity.setProperty(PROPERTY_NAME, this.name);
+    entity.setProperty(PROPERTY_DESCRIPTION, this.description);
+    entity.setProperty(PROPERTY_TASK_IDS, this.taskIds);
+
+    EmbeddedEntity userIdsEntity = new EmbeddedEntity();
+    for (Long key : this.userIds.keySet()) {
+      userIdsEntity.setProperty(key.toString(), this.userIds.get(key).name());
     }
+    entity.setProperty(PROPERTY_USER_IDS, userIdsEntity);
```
godsfavours, line 85: I wonder if there's another way to store a map in datastore w/o having to be concerned with parsing it when retrieving. It's straightforward with a set as seen in like 81, but not so much with map. EmbeddedEntity is the best way I found

src/main/webapp/index.html
```diff
@@ -6,7 +6,7 @@
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
-  <body>
+  <body onload="test()">
```
godsfavours: this has been removed along with the addition in script.js

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,208 @@
...
+      }
+    }
+  }
+
+  /**
+  * @return the entity representation of this class 
+  */
+  public Entity toEntity() {
+    Entity entity = new Entity("Project");
+    entity.setProperty(PROPERTY_NAME, this.name);
```
jcespejo, line 77: Smart use of constants for field names.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,208 @@
...
+ */
+package com.rtb.projectmanagementtool.project;
+
+import com.google.appengine.api.datastore.EmbeddedEntity;
+import com.google.appengine.api.datastore.Entity;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+
+enum User {
```
jcespejo, line 13: I would suggest renaming this to clarify the difference between this and the UserData object.  From the looks of it, something like "UserProjectRole" might make sense

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -0,0 +1,208 @@
...
+    Object userIdProperty;
+    if ((userIdProperty = entity.getProperty(PROPERTY_TASK_IDS)) != null) {
+      this.taskIds = new HashSet<Long>((ArrayList<Long>) userIdProperty);
+    } else {
+      this.taskIds = new HashSet<Long>();
+    }
+
+    this.userIds = new HashMap<Long, User>();
+    // can store map in datastore entities using EmbeddedEntity
+    EmbeddedEntity ee = (EmbeddedEntity) entity.getProperty(PROPERTY_USER_IDS);
```
jcespejo, line 66: Crafty usage of EmbeddedEntity to store maps.  Could be worth moving this into its own MapToEmbeddedEntity() method in a DataUtils.java file so everyone else can reuse.

Just keep in mind though that these are not searchable in a query.  Meaning you won't be able to do "get me projects owned by user X"

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
jcespejo: You can store a json string representing the whole map.  Or a Repeated property with json strings per item.  But either way you'd still need to parse (maybe json has an advantage because of built in parsing/encoding methods).  Storing them as an EmbeddedEntity I think is ok as well.

src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
```diff
@@ -0,0 +1,36 @@
...
+  public void getProjectsWithUser() {}
+
+  @Ignore
+  public void addProjects() {}
+
+  @Ignore
+  public void removeProject() {}
+
+  @Ignore
+  public void saveProjects() {}
```
jcespejo, line 34: Good approach stubbing out methods.

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Checks if the Project entity has at least one userId contained in userIds
+   *
+   * @param entity the project entity
+   * @param userIds the collection of userIds
+   * @return true if userIds param and entity's userIds property share at least one user
+   */
+  private boolean entityContainsUser(Entity entity, Collection userIds) {
```
jcespejo, line 56: I think the concept of 'Entity' we should try to contain within the Data object.  I see what you're doing here is for checking if a user exists on a project.  As an alternative, you can go ahead and build your ProjectData (based off of the Entity) then do checks off of the ProjectData instead of the entity.

Better approach would be to filter the query itself so we don't pull down all projects every time.  But I think that'll be tricky to do if we're storing a map of {role, userID}

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
godsfavours: I think I implemented the query functionality (unless I misunderstand what you mean by it) in getProjects(). It takes a Collection of userIds to fetch projects that only them are involved in. I could probably tweak the Collection type and my entityContainsUser() method to allow it to only grab projects owned by a user.

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
godsfavours: Yeah, makes sense. I sent in the entity so that I wouldn't have to actually create the ProjectData object before asserting that it's okay to add to the array returned by getProjects(). If it was created and entityContainsUser() returns false it would then just be garbage in the heap. What might it look like in java code to send in a query that would need to be filtered?

jcespejo: Datastore has a [filter functionality](https://cloud.google.com/appengine/docs/standard/java/datastore/queries#filters) so you wouldn't have to load all projects into memory (basically letting datastore do the heavy lifting for the filtering/sorting piece).  Although I don't think EmbeddedEntities are supported for filtering.

I think I've read somewhere that repeated fields can be filtered?  If this is the case, one approach would be to store a repeated list of strings in the format `{userID: role}` (ie json key:value format) so you can maybe do something like:
```
Filter projectAdminFilter =
    new FilterPredicate("users", FilterOperator.EQUAL, "{12345: 'ADMIN'}");
```

PR #32, opened 2020-07-28
Testing for UserController

talgsternberg:


src/main/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -0,0 +1,56 @@
...
+        HashSet<String> user1Majors = new HashSet();
+        HashSet<String> user2Majors = newHashSet();
+        user1Majors.add("Biology");
+        user2Majors.add("English");
+        user2Majors.add("Religion");
+        // Add 2 initially
+        ctl.addUserData(
+            ds, /*userID=*/ 1l, /*AuthID=*/ 2l,
+            /*userName=*/ "Sarah", /*userYear=*/ 2022,
+            /*userMajors=*/ user1Majors, /*skills=*/ Skills.LEADERSHIP,
```
talgsternberg: Skills.LEADERSHIP is what I found to access the enum vals online but doesn't seem to work. Any suggestions?

jcespejo: Your test is under the wrong package `package com.rtb.projectmanagementtool.taskblocker;`, try putting in the correct package so it'll find the Skills enum

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -42,12 +42,12 @@ public void testAddUsers() {
         ctl.addUser(
             ds, /*userID=*/ 1l, /*AuthID=*/ 2l,
             /*userName=*/ "Sarah", /*userYear=*/ 2022,
-            /*userMajors=*/ user1Majors, /*skills=*/ Skills.LEADERSHIP,
+            /*userMajors=*/ user1Majors, /*skills=*/ Skills.LEADERSHIP.name(),
```
talgsternberg, line 45: [ERROR] /home/talsternberg/project-management-tool/src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java:[45,76] incompatible types: java.lang
.String cannot be converted to com.rtb.projectmanagementtool.user.Skills

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -42,12 +42,12 @@ public void testAddUsers() {
...
             /*userTotalCompTasks=*/ 3);
         ctl.addUser(
             ds, /*userID=*/ 2l, /*AuthID=*/ 3l,
             /*userName=*/ "Joe", /*userYear=*/ 2024,
-            /*userMajors=*/ user2Majors, /*skills=*/ Skills.ORGANIZATION,
+            /*userMajors=*/ user2Majors, /*skills=*/ Skills.ORGANIZATION.name(),
```
talgsternberg, line 50: [ERROR] /home/talsternberg/project-management-tool/src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java:[50,78] incompatible types: java.lang
.String cannot be converted to com.rtb.projectmanagementtool.user.Skills

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
Tzirliya: I think you can remove the `.name()`.

talgsternberg: When I eliminate .name(), I get a test failure saying: "Tests in error:
  testAddUsers(com.rtb.projectmanagementtool.user.UserControllerTest): skills: com.rtb.projectmanagementtool.user.Skills is not a supported property type."

src/main/java/com/rtb/projectmanagementtool/user/UserController.java
```diff
@@ -24,4 +24,16 @@ public UserController(HashSet<UserData> users) {
     }
     return users;
   }
+
+  public void addUser(
+      DatastoreService datastore, long userID,
+      long AuthID, String userName,
```
jcespejo, line 30: My suggestion, controller shouldn't know too much about specifically what fields UserData owns (ie. what the data model looks like exactly).  Some fields would be required by the controller (IDs for example for doing modify/delete operations), but the other fields (username, year, major, etc..) the controller doesn't really need them.

Alternative approach would be to build your UserData outside of controller, and pass in the whole UserData instance into addUser.  This way you can change the data model (eg. add or remove fields) without having to touch your controller.

```
ctr.addUser(datastore, new UserData(userID, authID, userName, etc..))
```

(I believe my TaskBlocker example passes the individual fields, I apologize that was a lazy example)

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
jcespejo: This is probably from line 61 on your `toEntity` method.  skills at that point is still the enum, which Datastore does not support.  You'll need to convert to string when building your Entity, so Datastore can store it as a string.

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -40,15 +40,11 @@ public void testAddUsers() {
         user2Majors.add("Religion");
         // Add 2 initially
         ctl.addUser(
-            ds, /*userID=*/ 1l, /*AuthID=*/ 2l,
-            /*userName=*/ "Sarah", /*userYear=*/ 2022,
-            /*userMajors=*/ user1Majors, /*skills=*/ Skills.LEADERSHIP.name(),
-            /*userTotalCompTasks=*/ 3);
+            ds, new UserData(1l, 2l, "Sarah", 2022,
```
talgsternberg, line 43: testAddUsers(com.rtb.projectmanagementtool.user.UserControllerTest): java.lang.Long cannot be cast to java.lang.Integer

I can't quite figure out which param I passed in this is coming from ...

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -39,12 +39,18 @@ public void testAddUsers() {
...
         // Add 2 initially
-        ctl.addUser(
+        /* ctl.addUser(
             ds, new UserData(1l, 2l, "Sarah", 2022,
             user1Majors, Skills.LEADERSHIP.name(), 3));
         ctl.addUser(
             ds, new UserData(2l, 3l, "Joe", 2024,
-            user2Majors, Skills.ORGANIZATION.name(),5));
+            user2Majors, Skills.ORGANIZATION.name(),5)); */
+        ctl.addUser(
```
talgsternberg, line 48: This test still fails because "java.lang.Long cannot be cast to java.lang.Integer". I've tried casting each as it's respective type (long, int) in this assignment. I've also looked at UserController.java and UserData.java to ensure that I didn't use the incorrect order or mix up types there. Any tips/advice for how to proceed w/ this error?

src/main/java/com/rtb/projectmanagementtool/user/UserData.java
```diff
@@ -1,7 +1,7 @@
 package com.rtb.projectmanagementtool.user;
 
 import com.google.appengine.api.datastore.Entity;
-import java.util.HashSet;
+import java.util.*;
```
jcespejo, line 4: I would suggest listing out which specific utils you need from here.  I think java.util is huge and including all might not be efficient

src/main/java/com/rtb/projectmanagementtool/user/UserData.java
```diff
@@ -19,21 +19,21 @@
...
   private long AuthID; // this will be the ID from API
   private String userName;
-  private int userYear;
-  private HashSet<String> userMajors;
+  private long userYear;
+  private ArrayList<String> userMajors;
   // private Skills skills;
   // private String skillsString;
-  private int userTotalCompTasks;
+  private long userTotalCompTasks;
```
jcespejo, line 26: total number of completed tasks?  This could be ok, but keep in mind when they complete a task you'd now then have to update both the task object and this user object.

src/main/java/com/rtb/projectmanagementtool/user/UserController.java
```diff
@@ -12,8 +12,8 @@ public UserController(HashSet<UserData> users) {
   }
 
   // get users+data to direct/display user's profiles later
-  public HashSet<UserData> getEveryUser() {
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  public HashSet<UserData> getEveryUser(DatastoreService datastore) {
+    // DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     HashSet<UserData> users = new HashSet<>();
```
jcespejo, line 17: Watch out for variable scoping.  This `users` is a local variable, and not the class member (not `this.users`).  This is not necessarily bad though.  Maybe if you don't have a good use case for holding on to a class member, consider removing.

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import java.util.Arrays;
+import java.util.HashSet;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class UserControllerTest {
+  private HashSet<UserData> testUsers;
```
jcespejo, line 18: The way that this is setup, it looks like it'll be a list of users that's accessible throughout the whole test class.  For building test cases, it would be best to avoid shared resources between the different test methods if possible.  This means if you have test `testAddUsers` that adds items into `testUsers`, then later on you have some other test (let's say 'loadUsersTest') that looks at `testUsers`, you'd end up affecting your `loadUsersTest` test with results from your `testAddUsers` test.

pom.xml
```diff
@@ -59,6 +59,18 @@
           <deploy.version>1</deploy.version>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.coveo</groupId>
+        <artifactId>fmt-maven-plugin</artifactId>
+        <version>2.10</version>
```
talgsternberg, line 65: This is the correct place to change version, right?

PR #35, opened 2020-07-28
Tzirliya development

Tzirliya:
Some unit tests are added to test TaskData and TaskController. (More will be added in the future.) Many small bugs were found and fixed in the task classes.


src/main/java/com/rtb/projectmanagementtool/task/TaskData.java
```diff
@@ -42,17 +43,29 @@ public TaskData(Entity entity) {
...
     name = (String) entity.getProperty("name");
     description = (String) entity.getProperty("description");
-    status = (Status) entity.getProperty("status");
-    users = (HashSet<Long>) entity.getProperty("users");
-    subtasks = (HashSet<Long>) entity.getProperty("subtasks");
+    status = Status.valueOf((String) entity.getProperty("status"));
+    try {
+      users = new HashSet<Long>((ArrayList<Long>) entity.getProperty("users"));
+    } catch (ClassCastException e) {
+      users = (HashSet<Long>) entity.getProperty("users");
```
jcespejo, line 50: I'm curious about which scenarios where the first one would fail, but the second would be ok?  IIUC the datastore schema would hold the same data type (unless if you change), which I think would mean there wouldn't be a scenario where the same Property would be 2 different types (an ArrayList and a HashSet)?

src/main/java/com/rtb/projectmanagementtool/task/TaskData.java
```diff
@@ -113,4 +126,39 @@ public void setUsers(HashSet<Long> users) {
   public void setSubtasks(HashSet<Long> subtasks) {
     this.subtasks = subtasks;
   }
+
+  @Override
+  public int compareTo(TaskData task) {
+    long dif = this.taskID - task.taskID;
```
jcespejo, line 132: nit:  you can cast here just to avoid having to do the next line.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -16,14 +16,13 @@
 /** Class controlling the TaskData object. */
 public final class TaskController {
 
-  private HashSet<TaskData> tasks;
+  private DatastoreService datastore;
```
jcespejo, line 19: Good move removing the `tasks` member since the controller doesn't currently need to keep track of state anyways. (It can be added later on if there's a good use case for holding on to it).

I would suggest moving the Datastore ownership out of this class though, or at least initialize it with a copy from outside.  For some database engines, each instance of these would normally hold on to a connection to the database.  Meaning multiple instances would create multiple open connections.  Normally it would be best practice for applications to hold on to one open connection and share it throughout the application (by passing it around as params).  This saves resources by minimizing unneeded open connections.  It also ensures all usage are connected to the same exact database (in cases where you have multiple databases available to connect to).

src/test/java/com/rtb/projectmanagementtool/task/TaskTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+  @Test
+  public void testGetTasksFromEmptyDs() {
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+
+    // Assert no task entities are found
+    Assert.assertEquals(0, ds.prepare(new Query("Task")).countEntities(withLimit(10)));
+
+    // Get task entities with TaskController
+    TaskController taskController = new TaskController();
```
jcespejo, line 182: I think this helps illustrate why moving the Datastore out of TaskController could be good.  The way you have it here right now, the test in line 179 is using a different instance of Datastore than the one that's inside of TaskController.  In this instance they both connect to the same database (since the API takes care of that detail), but if you were using a different database engine, it could be possible for the one DS to connect to an entirely different database than the other.  (A common example would be where one connects to a Test database and the other connects to a Prod database).

src/test/java/com/rtb/projectmanagementtool/task/TaskTest.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/** Class testing tasks */
+public class TaskTest {
```
jcespejo, line 20: nit:  I would suggest separating the TaskData tests from the TaskController tests

src/main/java/com/rtb/projectmanagementtool/task/TaskData.java
Tzirliya: It's because while testing, I would create an entity from scratch using HashSets and transform it to a TaskData object without putting it on datastore. Therefore, it wouldn't become an ArrayList and would remain a HashSet. Would it be better to change the test so that an entity was created with an ArrayList? Or, should I put the entity in datastore first so that datastore would transform it for me? Or, should I leave it the way it is, with the TaskData class taking care of either option?

jcespejo: I see what you mean.  Generally you'd want your test to behave as close to prod behavior as possible.  If prod has entities with ArrayLists on them, you should also put together your tests to work with ArrayLists.  This will allow you to exercise your list->set conversion.  It'll also prevent you from having test-only code in your prod code.

Alternative approach would be to switch to ArrayList all together. This will make your Data more closely mimic what the actual DataStore data looks like.  Although you do lose the convenience of quickly looking up individual values from the set.  If you see yourself having to lookup specific values from the list/set, it could be worth it to do this extra conversion to Set.  If most of the time though you'll be iterating the whole list, maybe keep it as a List then only convert to Set when needed.

Tzirliya: I think that the majority of actions performed on users and subtasks would be iterating, adding, and removing. In that case, an ArrayList would probably be easier to work with overall.

jcespejo: I'm curious, can you elaborate a bit more on your thoughts regarding set vs. list?  (convenience, performance, efficiency, storage, etc..)

PR #36, opened 2020-07-29
Add TaskController Methods and Unit Tests

Tzirliya:
HashSets are changed to ArrayLists. `getTaskByID()` and `getSubtasks()` are added to TaskController. Unit tests are cleaned up. More unit tests are added.


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -22,8 +13,29 @@ public TaskController(DatastoreService datastore) {
     this.datastore = datastore;
   }
 
-  public HashSet<TaskData> getTasks(int quantity, String sortBy, String sortDirection) {
-    HashSet<TaskData> tasks = new HashSet<>();
+  public TaskData getTaskByID(long taskID) {
+    Query query = new Query("Task").addFilter("taskID", FilterOperator.EQUAL, taskID);
+    PreparedQuery results = datastore.prepare(query);
+    Entity entity = results.asIterable().iterator().next();
```
jcespejo, line 19: There's a `asSingleEntity` method that could work here as well.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -22,8 +13,29 @@ public TaskController(DatastoreService datastore) {
...
+    TaskData task = new TaskData(entity);
+    return task;
+  }
+
+  public ArrayList<TaskData> getSubtasks(TaskData task) {
```
jcespejo, line 24: This method and `getTasks` are mostly the same, just with different query options and filters.  Maybe look into extracting the common pieces into a `private` util method.

src/test/java/com/rtb/projectmanagementtool/task/TaskControllerTest.java
```diff
@@ -61,6 +67,58 @@ public void tearDown() {
...
+    // Assert task retrieved is correct
+    Assert.assertEquals("getTask", task2, getTask);
+  }
+
+  @Test
+  public void testGetSubtasks() {
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+
+    // Subtasks of task1
+    ArrayList<Long> subtasksParameter = new ArrayList<>(Arrays.asList(2l, 3l));
```
jcespejo, line 98: nit:  we don't need to tag as a "parameter".  If this was used later on for a non-param usages, it won't be accurate anymore.

src/test/java/com/rtb/projectmanagementtool/task/TaskControllerTest.java
```diff
@@ -134,16 +157,47 @@ public void testGetTasksFromDs() {
...
-    // Assert all 3 entities were retrieved
-    for (int i = 0; i < 3; i++) {
-      Assert.assertTrue("getTask", getTasksArrayList.get(i).equals(tasks.get(i)));
-    }
+    // Assert 3 entities were added
+    Assert.assertEquals(3, ds.prepare(new Query("Task")).countEntities(withLimit(10)));
+
+    // Delete task entities with TaskController
+    TaskController taskController = new TaskController(ds);
+    taskController.deleteTasks(new ArrayList<>(Arrays.asList(taskID1, taskID2, taskID3)));
```
jcespejo, line 198: It could be more interesting to see this delete some, but not all records.  It'll more clearly show if the code was misconfigured and deletes all records.

src/main/java/com/rtb/projectmanagementtool/task/TaskData.java
```diff
@@ -39,35 +38,36 @@ public TaskData(
...
+    if (entity.getProperty("subtasks") == null) {
+      subtasks = new ArrayList<Long>();
+    } else {
+      subtasks = (ArrayList<Long>) entity.getProperty("subtasks");
     }
   }
 
   public Entity toEntity() {
     Entity entity = new Entity("Task", taskID);
+    entity.setProperty("taskID", taskID);
```
jcespejo, line 60: What's the plan for generating taskID when inserting new records into datastore?

Tzirliya: It's probably not the best plan, but in the server I'll be creating an entity without the id parameter, so the entity will auto-generate an id. Then, I'll get the id from the entity and use that to create a TaskData object. The entity created before will never actually be stored in datastore and will only be used to create the id.

jcespejo: I see.  It might be a good idea to update your [controller test](https://github.com/googleinterns/step171-2020/blob/master/src/test/java/com/rtb/projectmanagementtool/task/TaskControllerTest.java#L167) to test for this assumption.  The expectation would be that each of the returned tasks would have their own unique IDs, and doing a "get task by ID" operation would give you back the task with that unique ID.

PR #39, opened 2020-07-30
failing test for UserDataTest

talgsternberg:


src/test/java/com/rtb/projectmanagementtool/user/UserDataTest.java
```diff
@@ -0,0 +1,114 @@
...
+    Assert.assertEquals("userID", userID1, user1.getUserID());
+    Assert.assertEquals("AuthID", AuthID1, user1.getAuthID());
+    Assert.assertEquals("userName", name1, user1.getUserName());
+    Assert.assertEquals("userYear", year1, user1.getUserYear());
+    Assert.assertEquals("userMajors", majors1, user1.getUserMajors());
+    Assert.assertEquals("userTotalCompTasks", total1, user1.getUserTotal());
+  }
+
+  @Test
+  public void testCreateFromEntity() {
```
talgsternberg, line 68: test failure: testCreateFromEntity(com.rtb.projectmanagementtool.user.UserDataTest): userID expected:<2> but was:<0>

Any ideas as to why this is the case?

godsfavours: An entity's key isn't initialized until its added to datastore w/ datastore.put(entity). Before that's done entity.getKey().getId() will always return 0. You've set the entity's id variable in the test, but when the user object is created its id is set to the entity's key

talgsternberg: Finally caught that. Thank you!

src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
```diff
@@ -0,0 +1,56 @@
...
+
+@WebServlet("/user")
+public class UserServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Query query = new Query("User");
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
```
Tzirliya, line 23: What's the purpose of this query?

jcespejo:
> Looks good. I just think we should figure out what we're doing with the JavaScript file. Are we sharing it or should we split it up for different pages? Sharing it would mean more merge conflicts and possibly more confusing code. Splitting it up would be easier to work with, but there might be more duplication because some pages can share the same functions.

Another option would be to separate into different files, then import one file into the other to be able to share reusable code.

PR #40, opened 2020-07-30
Add Task Servlet

Tzirliya:
A Task Servlet is added. The doGet() method returns a default task. The commented out code in the servlet is a general idea of what the servlet will look like once datastore is filled and all controller classes are completed. The getTaskInfo() function in the JavaScript file is modified to fetch from the servlet. The commented out code in that function is also a general idea of what the function will look like once all the pieces come together.


src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    // data.put("project", projectInArrayList);
+    // data.put("subtasks", subtasks);
+    // data.put("comments", comments);
+    // data.put("users", users);
+    response.getWriter().println(gson.toJson(data));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Entity entity = new Entity("Task"); // Used to generate taskID, but it's never put in datastore
```
jcespejo, line 72: Is this something you can encapsulate in the Data layer?  ie "if this TaskData is new, generate a new ID.  else use the ID this task is already holding on to"

jcespejo:
For the js file, also suggest to think about separating into multiple js files, or organizing your methods into classes.

Tzirliya:
I'm not sure if separating the js file makes most sense because some functions can be shared between pages. E.g. the getTasks() function can be used by both the task page (to show its subtasks) and the project page (to show it's tasks), and it can even be used by the user profile page possibly. The same thing can be done with a getUsers() function for the task and project pages.

src/test/java/com/rtb/projectmanagementtool/task/TaskServletTest.java
```diff
@@ -0,0 +1,86 @@
...
+    response = mock(HttpServletResponse.class);
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testDoGet() throws IOException, ServletException {
```
jcespejo, line 44: This test looks good.  Just a couple suggestions for the next PR:
-  If we're expecting JSON results, you can maybe parse the JSON into an object.  It doesn't have to be a TaskData object, but maybe a gson object would suffice.  This will allow us to assert the whole object, which would compare the results field by field.  Also if the order of the fields in the json string changes, asserting the object itself would allow for this
- If you'd rather keep the assert with a json string, look into formatting as a "raw string literal" or a "multi-line string literal" to improve readability, [eg](https://dzone.com/articles/jdk-12-raw-string-literals)
- Look into mocking the data for the test.  Either if you can wrap the datastore service with mockito (I'm not sure if this is possible), or by passing a local datastore into your servlet like how we did for the controller.  This allows for 2 things:  1.  you can control the output of the datastore and 2.  it wouldn't affect the production data.

src/test/java/com/rtb/projectmanagementtool/task/TaskServletTest.java
```diff
@@ -0,0 +1,86 @@
...
+    when(request.getParameter("name")).thenReturn("Task 1");
+    when(request.getParameter("description")).thenReturn("Task 1 description...");
+    when(request.getParameter("status")).thenReturn("incomplete");
+
+    // Create writer
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter printWriter = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(printWriter);
+
+    // Run doGet()
```
jcespejo, line 83: nit: Comment

src/test/java/com/rtb/projectmanagementtool/task/TaskServletTest.java
```diff
@@ -0,0 +1,86 @@
...
+    servlet.doPost(request, response);
+  }
```
jcespejo, line 85: What do we expect to happen after a successful post?  Is that something we can verify here?

PR #41, opened 2020-07-31
UserServlet 

talgsternberg:


src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
```diff
@@ -0,0 +1,67 @@
...
+    // long userYear1 = 2023;
+    // ArrayList<String> userMajors1 = new ArrayList<>(Arrays.asList("Geology", "Physics"));
+    // long userTotalCompTasks1 = 8;
+    // UserData user1 =
+    // new UserData(userID1, AuthID1, userName1, userYear1, userMajors1, userTotalCompTasks1);
+
+    // get the User
+    long userID = Long.parseLong(request.getParameter("userID"));
+    UserController userController = new UserController(datastore);
+    UserData user = UserController.getUserByID(userID);
```
talgsternberg, line 34: "non-static method getUserByID(long) cannot be referenced from a static context": I think I understand the difference between static/non static from some research I did this morning but I'm still a bit confused. getUserByID is defined in UserController and is only called in the server AFTER the UserController has been declared. Isn't this technically the structure that would allow for this call?

src/main/webapp/user_profile.html
```diff
@@ -47,6 +47,13 @@ <h2>User Checklist:</h2>
           <button class="complete" type="button" onclick="goToTask(4)">Task 4</button>
         </li>
       </ul>
+      <form method="GET" action="/user">
+        <button type ="submit" values="users" id="button" onClick="loadAllUsers() return false;">Get Users</button>
```
talgsternberg, line 51: What does it mean that "attribute names must follow format? I looked up some lint rules and it just said lower case with dashes if multi-word names. I'm not sure what has to change here ...

src/main/webapp/script.js
```diff
@@ -314,4 +314,30 @@ function createUserLiElement(user) {
...
+  const nameElement = document.createElement('span');
+  nameElement.innerText = user.userName;
+
+  userElement.appendChild(nameElement);
+  return userElement;
+}
+
+/** Fetches users from the server and adds them to the DOM. */
+function loadUsers() {
+  fetch('/user').then(response => response.json()).then((Users) => {
```
talgsternberg, line 336: Lint says I need extra parenthesis around arrow functions (ex (a) => {}), yet when I add extra parenthesis, I get errors saying "expected ;"

src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
jcespejo: ` UserData user = UserController.getUserByID(userID);`  the **U** in UserController is capital, so it would reference the class in a static context and not your actual local instance with the lower case **u**.

src/main/webapp/user_profile.html
jcespejo: recommended html format I believe is `attribute="value"` with no spaces before/after the equals sign.

src/main/webapp/script.js
jcespejo: `fetch('/user').then((response) => response.json())` or `fetch('/user').then((response) => (response.json()))` maybe?

src/main/webapp/user_profile.html
```diff
@@ -47,6 +47,13 @@ <h2>User Checklist:</h2>
           <button class="complete" type="button" onclick="goToTask(4)">Task 4</button>
         </li>
       </ul>
+      <form method="GET" action="/user">
+        <button type="submit" values="users" id="button" onClick="loadAllUsers()">Get Users</button>
```
talgsternberg, line 51: still the lint error: src/main/webapp/user_profile.html: line 51, col 58, attribute names must match the format

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -28,22 +62,34 @@ public void tearDown() {
     helper.tearDown();
   }
 
+  @Test
+  public void testGetUserByID() {
```
talgsternberg, line 66: testGetUserByID(com.rtb.projectmanagementtool.user.UserControllerTest)  Time elapsed: 0.033 sec  <<< ERROR!
java.lang.NullPointerException
        at com.rtb.projectmanagementtool.user.UserData.<init>(UserData.java:48)
        at com.rtb.projectmanagementtool.user.UserController.getUserByID(UserController.java:19)
        at com.rtb.projectmanagementtool.user.UserControllerTest.testGetUserByID(UserControllerTest.java:76)

talgsternberg: I'm having trouble working through this Null Pointer Exception. Can't seem to piece together exactly what'd going on here. Any glaring error/suggestions of how to work through?

src/main/java/com/rtb/projectmanagementtool/user/UserController.java
```diff
@@ -2,19 +2,27 @@
...
+
+  public UserController(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  public UserData getUserByID(long userID) {
+    Query query = new Query("User").addFilter("userID", FilterOperator.EQUAL, userID);
+    PreparedQuery results = datastore.prepare(query);
+    Entity entity = results.asSingleEntity();
+    UserData user = new UserData(entity);
```
jcespejo, line 19: The null pointer exception, this piece here is probably returning a null entity.  My guess is the query by ID didn't return a recrod, and asSingleEntity returned null.

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
Tzirliya: I would print the userData/entity along each step (in the test classes, userController methods used, and userData methods used) and pay attention to what the id is. Maybe something isn't what you expect it to be?

src/main/webapp/user_profile.html
Tzirliya: `onclick` is supposed to be all lowercase?

src/main/webapp/script.js
```diff
@@ -314,4 +314,30 @@ function createUserLiElement(user) {
   buttonElement.innerText = user.name;
   liElement.appendChild(buttonElement);
   return liElement;
+}
+
+/** Creates an element that represents a task, including its delete button.
+  * @param {Hashmap} user
```
emmacollins, line 320: should this also be Element?

src/main/webapp/script.js
```diff
@@ -314,4 +314,30 @@ function createUserLiElement(user) {
   buttonElement.innerText = user.name;
   liElement.appendChild(buttonElement);
   return liElement;
+}
+
+/** Creates an element that represents a task, including its delete button.
```
emmacollins, line 319: represents a user, instead of task?

src/main/webapp/script.js
```diff
@@ -314,4 +314,30 @@ function createUserLiElement(user) {
...
+  * @param {Hashmap} user
+  * @return {Element} HTML
+  */
+function createUserElement(user) {
```
emmacollins, line 323: what will createUserElement() be used for vs. createUserLiElement()?

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -12,7 +12,41 @@
...
 
 public class UserControllerTest {
+
+  // User 1
+  private static final long userID1 = 1l;
+  private static final long AuthID1 = 1l;
+  private static final String name1 = "Anna";
+  private static final long year1 = 2023;
+  private static final ArrayList<String> majors1 = new ArrayList<>(Arrays.asList("Biology", "Gov"));
+  private static final long total1 = 3;
```
emmacollins, line 22: nit: total what's this a total of?

talgsternberg: total completed tasks. I'll change it in my next PR.

src/main/webapp/script.js
talgsternberg: All of the js stuff to do with users is outdated. I have to rework it. I'll create another PR once I merge this one.

talgsternberg: But yes. I think it should be a UserData object?

src/main/webapp/script.js
talgsternberg: Not sure. Wrote the second one before checking to see if one was already written. My bad.

PR #43, opened 2020-08-01
Update Project classes

godsfavours:


src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
```diff
@@ -1,35 +1,136 @@
...
-  @Ignore
-  public void noProjects() {}
+  // One test method because I need tests to run in this order
+  @Test
+  public void runProjectControllerTest() {
+    testAddProjectsToDatabase();
+    testSaveProjectsToDatabase();
+    testGetProjectWithUserAndCreator();
+    testRemoveProjects();
+  }
```
Tzirliya, line 66: It would probably be better to keep the tests split up though because unit tests are supposed to be independent of one another. Maybe, for the tests that involve only retrieving some info from datastore, you can manually add entitities to datastore in that unit test without using projectController to add them? That way, you're only testing one function of projectController at a time.

jcespejo: +1.  tests requiring a specific order is usually discouraged.  it'll be hard to debug if the test fails mid way since you can't easily tell if the current test has problems, or if one of the earlier tests provided an incorrect state.

Also as a side note, it's advisable to "test behaviors and not methods/functions", so we don't necessarily need 1 test per function.  Most of the time they're synonymous if the method only performs a single behavior.  But for cases where a method does multiple behaviors, it would be good to have multiple tests for this same method (one per distinct behavior).  Or cases where multiple functions are used to perform 1 full action, it would be better to test all those methods in conjunction.

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -26,60 +22,70 @@ public ProjectController(DatastoreService datastore) {
...
+   * Gets all the projects in datastore that all users are in
    *
    * @param userIds the collection of userIds whose projects are to be retrieved
    * @return HashSet containing desired projects
    */
-  public HashSet<ProjectData> getProjects(Collection userIds) {
-
-    HashSet<ProjectData> projectContainer = new HashSet<ProjectData>();
+  public ArrayList<ProjectData> getAllProjects() {
+    return getProjects(0l, false);
```
jcespejo, line 31: Good use of a wrapper method.  I would suggest creating a named constant for the `0l` though (maybe `ANY_USER_ID`).  Also please add comments to specify what the hard coded params are, eg:
```
return getProjects(/*userId=*/ ANY_USER_ID, /*getCreatorProjects=*/false);
```

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -26,60 +22,70 @@ public ProjectController(DatastoreService datastore) {
...
+  }
 
-    return projectContainer;
+  /**
+   * Gets all the projects in datastore created by user
+   *
+   * @param userId the id of owner whose projects to retrieve
+   * @return HashSet containing desired projects
+   */
+  public ArrayList<ProjectData> getCreatorProjects(long userId) {
```
jcespejo, line 50: nit:  `Creator` in this form is an adjective of `Project`, whereas the intention here is for `Creator` to describe the type of user.  Maybe `getProjectsCreatedByUser` or `getProjectsWithCreator`

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -26,60 +22,70 @@ public ProjectController(DatastoreService datastore) {
...
-        projectContainer.add(project);
-      }
-    }
+  /**
+   * Gets all the projects in datastore that the specified user is in
+   *
+   * @param userId the id of user whose projects to retrieve
+   * @return HashSet containing desired projects
+   */
+  public ArrayList<ProjectData> getUserProjects(long userId) {
```
jcespejo, line 40: See comment below for `getCreatorProjects`.  For this one, technically all projects are User projects.  Maybe instead something like `getProjectsWithUser` could work better

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -26,60 +22,70 @@ public ProjectController(DatastoreService datastore) {
...
+   * @param userId the id of user whose projects to retrieve
+   * @param getCreatorProjects boolean representing the user's role in project (owner or otherwise)
+   * @return HashSet containing desired projects
    */
-  private boolean entityContainsUser(Entity entity, Collection userIds) {
-    // userIds is empty, so get all projects in database
-    if (userIds.size() == 0) {
-      return true;
-    }
+  private ArrayList<ProjectData> getProjects(long userId, boolean getCreatorProjects) {
```
jcespejo, line 61: Since you have 3 different user roles, you can improve flexibility here by checking for both the userId and the user role.  Maybe also rename the method to `getProjectsForUserWithRole`

src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -26,60 +22,70 @@ public ProjectController(DatastoreService datastore) {
...
+    ArrayList<ProjectData> projectContainer = new ArrayList<ProjectData>();
+
+    PreparedQuery results = datastore.prepare(new Query("Project"));
```
jcespejo, line 64: I highly advise not pulling all projects in the data store if possible.  This would take a long time to execute and eat up a lot of memory in a real production environment.  If searching by user is a key operation for this table, I would advise reevaluating how you're storing these to be able to cleanly filter by userIDs.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -140,56 +126,120 @@ public void setDescription(String description) {
...
   /**
-   * Adds a regular user to the project
+   * Helper function for addRegularUser() and addAdminUser() to add any user to the project. Can
+   * only add regular/admin users
    *
    * @param userId id of the user to add
+   * @param isAdmin is the user an admin?
    */
-  public void addRegularUser(long userId) {
+  private void addUser(long userId, boolean isAdmin) {
```
jcespejo, line 153: Mayber pass the userrole enum instead of a boolean since you have 3 different types.  I understand the 3rd type being the `CREATOR` it wouldn't change and wouldn't need to be accounted for by addUser.  But in the future when you add a 4th type (maybe VIEWER or AUDITOR, etc..) you'd then have to either add more bools here or change this into an enum anyways.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -140,56 +126,120 @@ public void setDescription(String description) {
...
    * Removes a user from the project
    *
    * @param userId id of the user to remove
+   * @return true if operation is successful
    */
-  public void removeUser(long userId) {
-    this.userIds.remove(userId);
+  public boolean removeUser(long userId) {
+    // don't remove creator
+    if (userType(userId) != UserProjectRole.CREATOR.name()) {
```
jcespejo, line 174: If the CREATOR is special enough and doesn't act as the two other users (cannot be added or remove), consider moving the CREATOR into it's own createroUserID field.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -140,56 +126,120 @@ public void setDescription(String description) {
...
+    return userType(userId) == UserProjectRole.CREATOR.name();
+  }
+
+  /**
+   * Returns the type of the user
+   *
+   * @param userId the userId
+   * @return the type of user (CREATOR / ADMIN / REGULAR)
+   */
+  private String userType(long userId) {
```
jcespejo, line 237: nit:  verbs for methods since methods are actions

src/main/java/com/rtb/projectmanagementtool/project/UserProjectRole.java
```diff
@@ -0,0 +1,14 @@
...
+// ADMIN - user with more administrative capabilities within project
+// CREATER - the creator of the project. Does everything an admin can do
+// and more, including being able to add/remove admins and delete the project
+
+package com.rtb.projectmanagementtool.project;
+
+public enum UserProjectRole {
+  CREATOR,
+  ADMIN,
+  REGULAR;
```
jcespejo, line 13: nit:  I wonder if there's another term we can use besides "REGULAR"?  It's hard to reason about how regular is different from the other types, since you can also argue that the ADMIN could be a "regular admin", if that makes sense?

Couple options, maybe project "MEMBER", "PARTICIPANT", "SUBSCRIBER" (although we're not using a subscription design pattern here), "CONTRIBUTOR"

src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
```diff
@@ -1,35 +1,136 @@
...
+
+  // Test saving projects to database
+  // Projects added here are used in subsequent test methods
+  public void testAddProjectsToDatabase() {
+    // Add projects to database to use get projects later
+    projectController.addProject(PROJECT1);
+    Assert.assertEquals(1, projectController.getAllProjects().size());
+
+    projectController.addProject(PROJECT2);
+    Assert.assertEquals(2, projectController.getAllProjects().size());
```
jcespejo, line 76: You don't necessarily have to assert every time you add (or maybe you don't have to invoke 4 times).  Conceptually the 3rd and 4th addProject calls don't differ from the first 2 and I can't think of a benefit for doing them that much. 

Doing twice is a good practice though, the second addProject confirms the previously added items were not overwritten.

Maybe as an alternative:  add 2 -> assert -> add 2 more -> assert again

src/test/java/com/rtb/projectmanagementtool/project/ProjectDataTest.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  // Test project creation given name, desc, and creator
+  @Test
+  public void testCreateProject() {
```
jcespejo, line 57: nit:  I don't think we need to prefix all test methods with `test` since the @Test annotation already does that for us.

PR #44, opened 2020-08-03
UserServer Test and Implement Enums

talgsternberg:
All tests pass. Ready for review.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
+++ b/src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
@@ -49,14 +49,19 @@ public class UserServlet extends HttpServlet {
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     Entity entity = new Entity("User");
+    System.out.println("no error yet: 1");
 
     // params from request
     long userID = (long) entity.getKey().getId();
     long AuthID = Long.parseLong(request.getParameter("AuthID"));
     String userName = request.getParameter("userName").trim();
     long userYear = Long.parseLong(request.getParameter("userYear"));
+
+    // error is here
+    // request.getParameterValues(userMajors) is null
     ArrayList<String> userMajors =
         (ArrayList<String>) Arrays.asList(request.getParameterValues("userMajors"));
+
     Skills skills = Skills.valueOf(request.getParameter("skills").toUpperCase());
     long userTotal = Long.parseLong(request.getParameter("userTotalCompTasks"));
 
--- a/src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
+++ b/src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
@@ -19,23 +19,26 @@ public class UserServlet extends HttpServlet {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
     // Test user
-    // long userID1 = 1l;
-    // long AuthID1 = 1l;
-    // String userName1 = "Jake";
-    // long userYear1 = 2023;
-    // ArrayList<String> userMajors1 = new ArrayList<>(Arrays.asList("Geology", "Physics"));
-    // long userTotalCompTasks1 = 8;
-    // UserData user1 =
-    // new UserData(userID1, AuthID1, userName1, userYear1, userMajors1, userTotalCompTasks1);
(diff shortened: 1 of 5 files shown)


src/test/java/com/rtb/projectmanagementtool/user/UserServletTest.java
```diff
@@ -0,0 +1,87 @@
...
+    response = mock(HttpServletResponse.class);
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testDoGet() throws IOException, ServletException {
```
talgsternberg, line 44: testDoGet(com.rtb.projectmanagementtool.user.UserServletTest)  Time elapsed: 0.01 sec  <<< ERROR!
java.lang.NumberFormatException: For input string: "3l"

src/test/java/com/rtb/projectmanagementtool/user/UserServletTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    String expectedResult =
+        "{\"user\":[{\"userID\":3l,\"AuthID\":3l,\"userName\":\"Sarah\",\"userYear\":2023,\"userMajors\":['Psychology},\"skills\":\"OOP\",\"userTotalCompTasks\":3]}";
+
+    // Assert results are as expected
+    Assert.assertEquals("doGet", result, expectedResult);
+  }
+
+  @Test
+  public void testDoPost() throws IOException, ServletException {
```
talgsternberg, line 68: testDoPost(com.rtb.projectmanagementtool.user.UserServletTest)  Time elapsed: 0.007 sec  <<< ERROR!
java.lang.NumberFormatException: For input string: "3l"

jcespejo: coming from HTTP (which is want mockito is mocking here), you don't need the "l" after the number.  The "l" forces java to create a literal `long` instead of an `int`.

src/main/java/com/rtb/projectmanagementtool/user/UserServlet.java
```diff
@@ -46,18 +49,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
jcespejo, line 50: This servlet handles "/user" requests. What exactly do we POST here?  I see add users operation below, but I believe there's also going to be an "update user" or maybe even a  "delete user" operation here.  We should think about how we would structure our servlets based on the different operations.  Either having multiple servlets (one per operation), or having a single servlet that takes an "operation" param.  I would suggest discussing with the team to come up with a common convention.

src/test/java/com/rtb/projectmanagementtool/user/UserServletTest.java
```diff
@@ -0,0 +1,89 @@
...
+    servlet.doGet(request, response);
+
+    // Get results of doGet()
+    String result = stringWriter.getBuffer().toString().trim();
+
+    String expectedResult =
+        "[{\"userID\":3,\"AuthID\":3,\"userName\":\"Sarah\",\"userYear\":2023,\"userMajors\":[\"Psychology\"],\"skills\":\"OOP\",\"userTotalCompTasks\":3}]";
+
+    // Assert results are as expected
+    Assert.assertEquals("doGet", result, expectedResult);
```
jcespejo, line 65: If the expected result is in json format, I would suggest parsing the json and validating against the json object.  This will help with scenarios where if the field order changes, validating as strings would breaks, whereas validating as json objects would be order-agnostic.

src/test/java/com/rtb/projectmanagementtool/user/UserServletTest.java
```diff
@@ -0,0 +1,89 @@
...
+    when(request.getParameter("userTotalCompTasks")).thenReturn("3");
+
+    // Create writer
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter printWriter = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(printWriter);
+
+    // Run doPost()
+    servlet.doPost(request, response);
+  }
```
jcespejo, line 88: What are we testing for here?  Conceptually if the POST does things, we should be able to validate those "things".

PR #45, opened 2020-08-04
Project Servlets and Tests

godsfavours:
Created Servlets for Projects as well as accompanying tests. PR also includes improvements based on comments on previous PR

### DIFF
--- a/src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
+++ b/src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
@@ -2,13 +2,13 @@ import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
 import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
-import com.rtb.projectmanagementtool.project.ProjectController;
-import com.rtb.projectmanagementtool.project.ProjectData;
+import com.rtb.projectmanagementtool.project.*;
 import java.util.ArrayList;
 import java.util.Arrays;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class ProjectControllerTest {
@@ -36,13 +36,13 @@ public class ProjectControllerTest {
     projectController = new ProjectController(ds);
 
     // Add users to projects
-    PROJECT2.addAdminUser(USER2);
-    PROJECT2.addRegularUser(USER4);
-    PROJECT2.addAdminUser(USER3);
-    PROJECT1.addAdminUser(USER4);
-    PROJECT4.addRegularUser(USER2);
-    PROJECT3.addRegularUser(USER3);
-    PROJECT3.addRegularUser(USER1);
+    PROJECT2.addUser(USER2, UserProjectRole.ADMIN);
+    PROJECT2.addUser(USER4, UserProjectRole.MEMBER);
+    PROJECT2.addUser(USER3, UserProjectRole.ADMIN);
+    PROJECT1.addUser(USER4, UserProjectRole.ADMIN);
+    PROJECT4.addUser(USER2, UserProjectRole.MEMBER);
+    PROJECT3.addUser(USER3, UserProjectRole.MEMBER);
+    PROJECT3.addUser(USER1, UserProjectRole.MEMBER);
 
     // Set hardcoded ids to mimic expected behavior
     PROJECT1.setId(1l);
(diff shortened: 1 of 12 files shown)


src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
```diff
@@ -56,81 +56,80 @@ public void tearDown() {
...
-    testSaveProjectsToDatabase();
-    testGetProjectWithUserAndCreator();
-    testRemoveProjects();
-  }
-
   // Test saving projects to database
   // Projects added here are used in subsequent test methods
-  public void testAddProjectsToDatabase() {
-    // Add projects to database to use get projects later
+  @Ignore
```
jcespejo, line 61: Why ignore?

src/test/java/com/rtb/projectmanagementtool/project/ProjectControllerTest.java
```diff
@@ -56,81 +56,80 @@ public void tearDown() {
...
-    Assert.assertEquals(3, projectController.getAllProjects().size());
+    ArrayList<ProjectData> expectedProjects =
+        new ArrayList<ProjectData>(Arrays.asList(PROJECT1, PROJECT2, PROJECT3, PROJECT4));
+
+    Assert.assertEquals(expectedProjects.toString(), projectController.getAllProjects().toString());
 
+    expectedProjects.remove(PROJECT2);
     projectController.removeProject(PROJECT2);
-    Assert.assertEquals(2, projectController.getAllProjects().size());
+    Assert.assertEquals(expectedProjects.toString(), projectController.getAllProjects().toString());
```
jcespejo, line 127: Junit should be able to assert the objects themselves, instead of converting to strings.  Maybe look into if you can assert these without writing to strings first.

src/test/java/com/rtb/projectmanagementtool/project/servlets/DeleteProjectServletTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    // Assert that the project was added
+    Assert.assertEquals(1, projectController.getAllProjects().size());
+
+    when(request.getParameter("id")).thenReturn("1");
+
+    // Run servlet
+    servlet.doPost(request, response);
+
+    Assert.assertEquals(Collections.emptyList(), projectController.getAllProjects());
```
jcespejo, line 56: Would this pull from the list of projects in prod?

src/test/java/com/rtb/projectmanagementtool/project/servlets/DeleteProjectServletTest.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void deleteProject() throws Exception {
+    projectController.addProject(new ProjectData("Project", "Desc", 0l));
```
jcespejo, line 46: Watch out with these since this will mean if your controller addProject breaks (or if you update the logic), your servlet tests could also break. One alternative you could look into is if we can use mockito to mock the controller, so we don't actually invoke the controller when testing the servlet.

src/test/java/com/rtb/projectmanagementtool/project/servlets/DisplayProjectsServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Test
+  public void displayProjectsWithUser() throws Exception {
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter printWriter = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(printWriter);
+    when(request.getParameter("query")).thenReturn("isIn");
+
+    // Create project - user should be in this one
+    ProjectData newProject = new ProjectData("Project1", "Desc", 2l);
```
jcespejo, line 54: If we're going to be adding multiple projects at a time, maybe look into expanding your controller to allow for adding a list of projects (to consolidate the datastore call into a single insert)

src/test/java/com/rtb/projectmanagementtool/project/servlets/DisplayProjectsServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+    // Run Servlet
+    servlet.doPost(request, response);
+
+    ArrayList<String> expectedOutput =
+        new ArrayList<String>(Arrays.asList(newProject.getName(), newProject3.getName()));
+    String actualOutput = stringWriter.getBuffer().toString().trim();
+
+    // Converting expectedOutput to gson because servlet returns stringWriter as json
+    Gson gson = new Gson();
+    Assert.assertEquals(gson.toJson(expectedOutput).toString(), actualOutput);
```
jcespejo, line 75: Same here, if possible please avoid asserting string conversions of the object, but instead assert on the object themselves.

PR #46, opened 2020-08-04
AuthServlet and Test Class

talgsternberg:


src/main/java/com/rtb/projectmanagementtool/user/AuthServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    String urlToRedirectToAfterUserLogsIn; // go to index (all user data should load)
+    String urlToRedirectToAfterUserLogsOut; // go to index with other pages blocked
+    String loginUrl;
+    String logoutUrl;
+
+    // get all the properties
+    if (userService.isUserLoggedIn()) {
+      status = "In";
+      userEmail = userService.getCurrentUser().getEmail();
+      AuthID = userService.getCurrentUser().getId();
```
talgsternberg, line 38: Apparently getId() isn't a valid method (got it from a tutorial but I think it was for something else now that I look at it again). If anyone has a tutorial or suggestions lmk  :)

jcespejo: `getUserId` https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User.html#getUserId--


I'm curious though how exactly we're planning on using this and how it connects to all the other servlets.  This might be worth explaining in the design doc before implementing

talgsternberg: Wrote more info on Design Doc.

talgsternberg: Lmk what you all think

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private DatastoreService datastore;
+  public String cookieName;
+  public String[] cookieValues;
+  public Cookie userCookie;
+
+  public AuthOps(DatastoreService datastore) {
+    this.datastore = datastore;
+    cookieName = "loginStatus";
+    cookieValues = {"out", "null"};
```
talgsternberg, line 17: Is this the right idea for storing the cookie vals? I researched it had to be a String so I thought a String array might work. Getting a lot of invalid expression errors.

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
```diff
@@ -0,0 +1,46 @@
...
+    if (userService.isUserLoggedIn()) {
+      
+      // set the first cookie value status to "in"
+      userCookie.getValue[0] = "in";
+      
+      // get AuthID
+      AuthID = userService.getCurrentUser().getUserId();
+      
+      // find AuthID in DataStore
+      Query query = new Query("User").addFilter("AuthID", FilterOperator.EQUAL, userID);
```
talgsternberg, line 37: At this point what's the best way to error if the user isn't found?

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
```diff
@@ -0,0 +1,46 @@
...
+public final class AuthOps {
+
+  private DatastoreService datastore;
+  public String cookieName;
+  public String[] cookieValues;
+  public Cookie userCookie;
+
+  public AuthOps(DatastoreService datastore) {
+    this.datastore = datastore;
+    cookieName = "loginStatus";
```
jcespejo, line 16: We're not just storing a single status here, since we said we'd also use this to track who is logged in.  So maybe "loggedInUserId" or "sessionUserId" would work instead

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
jcespejo, line 17: For the value, if we're storing the userId in our cookie, then we don't need multiple values. It'll either be the actual userId, or some constant to specify that the user is not  logged in (eg. zero or -1 are common).  We maybe also don't need to default the value here.

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
```diff
@@ -0,0 +1,46 @@
...
+    cookieValues = {"out", "null"};
+  }
+
+  public void loginUser() {
+    String AuthID;
+    userCookie = new Cookie(cookieName, cookieValues);
+
```
jcespejo, line 23: Try to follow the steps you have in the Detailed Design.  Step #1.a is saying "using cookies, c heck loggedin status", which means we'll first need to look at existing cookies to see if we already have loggedInUser tag in there.

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
jcespejo, line 37: Try to see if you can use your UserController to do this query, instead of putting together a new query here.

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
```diff
@@ -0,0 +1,46 @@
+package com.rtb.projectmanagementtool.user;
```
jcespejo, line 1: This would be a new feature that's related to User, but not exactly the same as User.  It would be good to put this in its own package.  Also naming wise, I would suggest "AuthHelper" or "AuthUtils".  "AuthOps" is also good, but the former 2 are more standard and more flexible if you need to add a non-operation method.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private DatastoreService datastore;
+  public String cookieName;
+  public String cookieValue;
+  public Cookie currCookie;
+
+  public AuthOps(DatastoreService datastore) {
+    this.datastore = datastore;
+    cookieName = "sessionUserID";
+    cookieValue = "out"; // "out" if not logged in
```
jcespejo, line 22: If this will normally store userIDs (which are ints), I would suggest also storing an int for your "NOT_LOGGED_IN" case.  This way you can easily cast from int<->string without having to worry about if it's a number or not.  You can do zero or -1.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public void loginUser(HttpServletRequest request) {
+
+    // get all cookies from request
+    Cookie[] cookies = request.getCookies();
```
jcespejo, line 28: Looks good.  You're going to be reusing this piece again later in your other methods, so it might be good to move this into it's own GetCookieFromRequest method.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,59 @@
...
+    // if not logged in, call auth service
+    if (currCookie.getValue() == "out") {
+      // call auth service
+      UserService userService = UserServiceFactory.getUserService();
+      if (userService.isUserLoggedIn()) {
+        // get AuthID
+        String AuthID = userService.getCurrentUser().getUserId();
+        // find AuthID in DataStore
+        UserController controller = new UserController(datastore);
+        ArrayList<UserData> users = controller.getEveryUser();
```
jcespejo, line 50: This is ok for now, but before you do your next method, maybe look into creating a `getUserWithAuthId` method in your controller.  See Tzirliya's question on the chat just now for info on how to do filters.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,59 @@
...
+        for (UserData user : users) {
+          if (user.getAuthID() == AuthID) {
+            currCookie.setValue(AuthID);
```
jcespejo, line 53: We would want to set the userID here, not the authID.  Since what we need the cookie for is to tell us which userID is logged in

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,59 @@
...
+          }
+        }
```
jcespejo, line 55: Missing the last step of sending the cookie back to the response.

We'll also need to take care of error scenarios, but that can be added later.

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testLoginUser() {
+    // new controller
+    UserController controller = mock(UserController.class);
+    controller.addUser(testUser);
```
jcespejo, line 59: We don't need to actually add the user into the datastore.  The reason we're mocking UserController is so we don't have to use a real UserController.

What we'll need instead is the `when().thenReturn()` line to have the controller return the test user when getUser is called

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    // build and send a test cookie
+    cookie = mock(Cookie.class);
+    cookie.setName("sessionUserID");
+    cookie.setValue("out");
+
```
jcespejo, line 65: We would also need to setup the when.then for the HTTPServletRequest.  We want it so **when** getCookies is called, **then** it returns our request_cookie

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -0,0 +1,71 @@
...
+    // call loginUser
+    loginUser(request, response);
+
+    Assert.assertEquals(cookie.getValue(), String.valueOf(userID));
```
jcespejo, line 69: This is the part I'm not sure if we need the "captor" setup.  Start off with:
`Assert.assertEquals(response.getCookies(), *THE_WE'RE_EXPECTING*)'

If that doesn't work, then look into setting up the captor

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -54,18 +43,35 @@ public void tearDown() {
...
+    // when get cookies is called, pass cookies
+    when(request.getCookies()).thenReturn(testCookies);
 
     // call loginUser
-    loginUser(request, response);
+    auth.loginUser(request, response);
+
+    // captor setup
+    ArgumentCaptor<Cookie> cookieCaptor = ArgumentCaptor.forClass(Cookie.class);
+    verify(response).addCookie(cookieCaptor.capture());
```
talgsternberg, line 72: Wanted but not invoked:
httpServletResponse.addCookie(
    <Capturing argument>
);

talgsternberg: I think I'm not understanding exactly what's going on here. I'm trying to get the cookies from the response (ie: response.getCookies(), but this way because that method doesn't exist. This line (72), I think is implementing a way to "get" the cookies from the response. Yet, there are zero interactions with this mock.

jcespejo: starting off with the "wanted but not invoked" error, that means we're expecting the test to call addCookie but it doesn't get called.  Most probably that [line in authHelpers](https://github.com/googleinterns/step171-2020/pull/46/files/6bf3c7946b97740c326ac923f70145bf118fe057#diff-975ee4eb190e78d0a053b7d66bafe0abR60) is not getting reached.  Maybe add a couple print statements in there to see why it's not getting to that last bit.

jcespejo: lines 71-75 I think is good

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -54,18 +43,35 @@ public void tearDown() {
...
-    controller.addUser(testUser);
 
     // build and send a test cookie
-    cookie = mock(Cookie.class);
-    cookie.setName("sessionUserID");
-    cookie.setValue("out");
+    Cookie[] testCookies = new Cookie[3];
+    testCookies[0] = new Cookie("SessionUserID", "abc");
+    testCookies[1] = new Cookie("SessionUserID", "zvm");
+    testCookies[2] = new Cookie("SessionUserID", "wpk");
```
jcespejo, line 59: If we're trying to test the login scenario, we would want this to look like there was no user that's logged in.  Either don't set SessionUserId, or set it to the NO_USER_LOGGED_IN value (which we were discussing could be zero or -1).

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -75,60 +83,43 @@ public void testAlreadyLoggedInDontSetCookie() {
 
   @Test
   public void testLoginUser() {
-
-    System.out.println("Checkpoint 1");
+    // auth object
+    ds = DatastoreServiceFactory.getDatastoreService();
```
talgsternberg, line 87: class DatastoreService, DataServiceFactory, ds not recognized and can't figure out why ... missing imports?

Tzirliya: Are you missing the type? `DatastoreService ds = ...`

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.mockito.*;
+
+public class AuthOpsTest {
+
+  private HttpServletRequest request;
+  private HttpServletResponse response;
+  private Cookie cookie;
+  private AuthOps auth;
+
+  private final LocalServiceTestHelper helper =
```
jcespejo, line 26:
```
private final LocalServiceTestHelper helper =
      new LocalServiceTestHelper(
              new LocalUserServiceTestConfig().setOAuthUserId("abc"),
              new LocalDatastoreServiceTestConfig()
                  .setAutoIdAllocationPolicy(
                      LocalDatastoreService.AutoIdAllocationPolicy.SEQUENTIAL))
          .setEnvIsLoggedIn(true);
```

jcespejo: method `setUserServiceAuthInfo()` you can call in your tests to change the login status for testing

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,85 @@
...
+
+  // private DatastoreService datastore;
+  private UserController controller;
+  public String cookieName;
+  public String cookieValue;
+  public Cookie currCookie;
+
+  public AuthOps(UserController controller) {
+    this.controller = controller;
+    cookieName = "sessionUserID";
```
jcespejo, line 22: These won't change.  These can be static constants for your whole class.  Also for cookieValue, since this specifically holds the "not logged in" value, try to rename the variable as such.  Check what naming convention the team is using for static variables.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class AuthOps {
+
+  // private DatastoreService datastore;
+  private UserController controller;
+  public String cookieName;
+  public String cookieValue;
+  public Cookie currCookie;
+
+  public AuthOps(UserController controller) {
```
jcespejo, line 20: As per discussion, this constructor is for testing only so we can pass a mock UserController in.  Please add a comment stating this.  Also please add a default constructor (with no params) that builds a non-test controller

PR #48, opened 2020-08-05
Change user storage in ProjectData

godsfavours:
Changed user storage in ProjectData and updated ProjectDataTest


src/test/java/com/rtb/projectmanagementtool/project/ProjectDataTest.java
```diff
@@ -54,123 +58,145 @@ public void tearDown() {
...
+    entity.setProperty(PROPERTY_ADMINS, new HashSet<Long>(Arrays.asList(USER3, USER4)));
+    entity.setProperty(PROPERTY_MEMBERS, new HashSet<Long>(Arrays.asList(USER1)));
 
     // Create project
     ProjectData project = new ProjectData(entity);
-    Assert.assertEquals(expectedUsers, project.getUsers());
+
+    // Have to use toString() because while entities have the same variables, they don't have the
+    // same references
+    Assert.assertEquals("project matches entity", entity.toString(), project.toEntity().toString());
```
godsfavours, line 96: Without toString():
```
Assert.assertEquals("project matches entity", entity, project.toEntity());
```
java.lang.AssertionError: project matches entity expected: com.google.appengine.api.datastore.Entity<<Entity [Project(no-id-yet)]:
        creator = 2
        members = [1]
        name = Project 1
        description = Description of Project 1
        tasks = []
        admins = [3, 4]
>
> but was: com.google.appengine.api.datastore.Entity<<Entity [Project(no-id-yet)]:
        creator = 2
        members = [1]
        name = Project 1
        description = Description of Project 1
        tasks = []
        admins = [3, 4]

jcespejo: Equality for two entities is true only if the key objects match:
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Entity.html#equals-java.lang.Object-

It sounds like it's not looking at the values itself, which is what you need.  Couple options you can do:  convert them as ProjectData and compare those (assuming you implemented the equals operation for your ProjectData).  Or create a reusable `CompareEntities(Entity e1, Entity e2)` method that extracts the properties using 'getProperties()' and compares them one by one

PR #50, opened 2020-08-05
Update Task Classes

Tzirliya:
Update Task Classes based on suggestions from the previous pull request.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
@@ -13,22 +13,29 @@ public final class TaskController {
     this.datastore = datastore;
   }
 
-  public ArrayList<Key> addTasks(ArrayList<TaskData> tasks) {
+  public void addTasks(ArrayList<TaskData> tasks) {
     ArrayList<Entity> taskEntities = new ArrayList<>();
     for (TaskData task : tasks) {
       taskEntities.add(task.toEntity());
     }
-    return new ArrayList<>(datastore.put(taskEntities));
+    addKeysToTasks(tasks, new ArrayList<>(datastore.put(taskEntities)));
   }
 
-  public ArrayList<Key> addSubtasks(TaskData task, ArrayList<TaskData> subtasks) {
-    ArrayList<Key> keys = addTasks(subtasks);
-    ArrayList<Long> taskSubtasks = task.getSubtasks();
-    for (Key key : keys) {
-      taskSubtasks.add(key.getId());
+  public void addSubtasks(TaskData task, ArrayList<TaskData> subtasks) {
+    addTasks(subtasks);
+    ArrayList<Long> subtaskIDs = task.getSubtasks();
+    for (TaskData subtask : subtasks) {
+      subtaskIDs.add(subtask.getTaskID());
     }
-    task.setSubtasks(taskSubtasks);
-    return keys;
+    task.setSubtasks(subtaskIDs);
+  }
+
+  public void addKeysToTasks(ArrayList<TaskData> tasks, ArrayList<Key> keys) {
+    if (tasks.size() == keys.size()) {
+      for (int i = 0; i < keys.size(); i++) {
+        tasks.get(i).setTaskID(keys.get(i).getId());
+      }
+    }
   }
 
   public TaskData getTaskByID(long taskID) {
@@ -92,6 +99,14 @@ public final class TaskController {
     ArrayList<Key> keys = new ArrayList<>();
     for (long taskID : taskIDs) {
       keys.add(KeyFactory.createKey("Task", taskID));
+    }
+    return keys;
+  }
+
+  public ArrayList<Key> getKeysFromTasks(ArrayList<TaskData> tasks) {
+    ArrayList<Key> keys = new ArrayList<>();
+    for (TaskData task : tasks) {
+      keys.add(KeyFactory.createKey("Task", task.getTaskID()));
     }
     return keys;
   }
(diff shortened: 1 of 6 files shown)


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -76,8 +77,12 @@ public TaskData getTaskByID(long taskID) {
...
 
   public void deleteTasks(ArrayList<Long> taskIDs) {
-    ArrayList<Key> keys = getKeysFromTaskIDs(taskIDs);
-    datastore.delete(keys);
+    if (taskIDs.isEmpty()) {
+      return;
+    }
+    datastore.delete(getKeysFromTaskIDs(taskIDs));
+    Query query = new Query("Task").addFilter("parentTaskID", FilterOperator.IN, taskIDs);
+    deleteTasks(getTaskIDsFromTasks(getTasks(query, Integer.MAX_VALUE)));
```
jcespejo, line 85: Is there a use case for limiting the number of records to pull?  

If you'd like to keep it, consider storing this on a better named constant like `NO_QUERY_LIMIT`, or maybe overload your method so you won't need to pass around MAX_VALUE if you need to pull all records.

src/test/java/com/rtb/projectmanagementtool/task/TaskControllerTest.java
```diff
@@ -66,111 +55,99 @@ public void tearDown() {
...
+        new Query("Task").addFilter("parentTaskID", FilterOperator.EQUAL, task1.getTaskID());
     PreparedQuery results = ds.prepare(query);
 
     // Assert correct subtasks were found
+    TaskData task;
     for (Entity entity : results.asIterable()) {
-      long taskID = entity.getKey().getId();
-      Assert.assertTrue("addSubtasks", subtaskIDs.contains(taskID));
+      task = new TaskData(entity);
+      Assert.assertTrue("addSubtasks", subtasks.contains(task));
```
jcespejo, line 103: assertEquals can do list of objects I believe, so you can do something like `assertEquals(expected_subtasks, actual_subtasks)` instead of looping through them 1 by 1.  It sounds like more work (since you'll have to loop once to build actual_subtasks, then presumably assertEquals would loop as well), but in error scenarios it would print more details info like "expected contains tasks 1, 2, 3;  actual contains tasks 2, 3", compared to assertTrue which would only report back "result is false"

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -76,8 +77,12 @@ public TaskData getTaskByID(long taskID) {
   }
 
   public void deleteTasks(ArrayList<Long> taskIDs) {
-    ArrayList<Key> keys = getKeysFromTaskIDs(taskIDs);
-    datastore.delete(keys);
+    if (taskIDs.isEmpty()) {
+      return;
+    }
```
jcespejo, line 82: Interesting approach with the recursive deletes.  So this calls getTasks recursively, then calls delete for each returned set.  One possible danger here would be if one of the deletes failed, you'd then be left with orphan records (since the parent task no longer exists).

A couple alternatives you can look  into:
1.  Still do recursive getTasks but do one giant delete at the end
2.  Look into enabling "transactions"

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -53,12 +57,9 @@ public TaskData getTaskByID(long taskID) {
   }
 
   public ArrayList<TaskData> getSubtasks(TaskData task) {
-    ArrayList<Key> subtaskKeys = getKeysFromTaskIDs(task.getSubtasks());
-    if (!task.getSubtasks().isEmpty()) {
-      Query query = new Query("Task").addFilter("__key__", FilterOperator.IN, subtaskKeys);
-      return getTasks(query, Integer.MAX_VALUE);
-    }
-    return new ArrayList<>();
+    Query query =
```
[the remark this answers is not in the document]
Tzirliya: I don't think `setOrderBy` and `setLimit` work with this version of datastore.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
Tzirliya: There's no use case now, but maybe there will be in the future (especially if we add a global search feature - to search through all tasks/projects visible to a user).

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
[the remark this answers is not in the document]
jcespejo: I see.  Looks like you'll still need to pass sort field/direction separately (or create a class to hold them together).  For limit, you can use FetchOptions:  https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/FetchOptions.html

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -41,45 +53,64 @@ public TaskData getTaskByID(long taskID) {
...
+
   public void deleteTasks(ArrayList<Long> taskIDs) {
-    ArrayList<Key> keys = getKeysFromTaskIDs(taskIDs);
-    datastore.delete(keys);
+    TransactionOptions options = TransactionOptions.Builder.withXG(true);
+    Transaction transaction = datastore.beginTransaction(options);
+    try {
+      deleteTasksRecursive(taskIDs);
+      transaction.commit();
+    } finally {
```
jcespejo, line 96: Should this be `catch` instead of `finally`?   iiuc finally is executed every time the try block exits.  Which means on successful operations, this would also rollback?

Tzirliya: In the documentation it used `finally`. Maybe the transaction isn't considered active after committing? It passes the unit tests though.

jcespejo: It could be possible it would auto close right after the commit.  In which case the finally would be ok.

PR #51, opened 2020-08-06
Updates to Project classes

godsfavours:
Admin and Member users, along with any new roles in the future, now stored in HashMap<UserProjectRole, HashSet<Long>> in java. HashSets of each member role are stored independently in datastore (admins have their own hashset, etc.). Thoughts on this approach?


src/main/java/com/rtb/projectmanagementtool/project/ProjectController.java
```diff
@@ -29,52 +31,45 @@ public ProjectController(DatastoreService datastore) {
...
 
+    ArrayList<FilterPredicate> subFilters = new ArrayList<FilterPredicate>();
+    for (UserProjectRole userRole : userRoles) {
+      if (userRole == UserProjectRole.CREATOR) {
+        subFilters.add(new FilterPredicate("creator", FilterOperator.EQUAL, userId));
+      } else if (userRole == UserProjectRole.ADMIN) {
+        subFilters.add(new FilterPredicate("admins", FilterOperator.IN, Arrays.asList(userId)));
+      } else if (userRole == UserProjectRole.MEMBER) {
+        subFilters.add(new FilterPredicate("members", FilterOperator.IN, Arrays.asList(userId)));
+      }
```
Tzirliya, line 57: You can use a switch case here.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -41,32 +52,57 @@ public ProjectData(String name, String description, long creatorId) {
...
   /** @return the entity representation of this class */
   public Entity toEntity() {
     Entity entity = new Entity("Project");
     entity.setProperty(PROPERTY_NAME, this.name);
+    entity.setProperty(PROPERTY_CREATOR, this.creatorId);
     entity.setProperty(PROPERTY_DESCRIPTION, this.description);
     entity.setProperty(PROPERTY_TASKS, this.tasks);
-    entity.setProperty(PROPERTY_USERS, this.users);
+    entity.setProperty(PROPERTY_ADMINS, this.users.get(UserProjectRole.ADMIN));
+    entity.setProperty(PROPERTY_MEMBERS, this.users.get(UserProjectRole.MEMBER));
```
Tzirliya, line 105: If admins and members are stored as two separate properties in an entity in datastore, is there a reason for combining them in the ProjectData class as a hashmap instead of having an admins set and a members set?

jcespejo:
> Admin and Member users, along with any new roles in the future, now stored in HashMap<UserProjectRole, HashSet> in java. HashSets of each member role are stored independently in datastore (admins have their own hashset, etc.). Thoughts on this approach?

Storing as multiple flat lists in datastore I think would be a good idea (as long as we don't have plans to add more user roles later on).  It definitely would make filtering by these IDs easier.  If we're doing it this way, we probably don't need to build the complex hashmap in java as well.  If you keep them as separate sets, it'll more closely match the data model, but you won't have to deal with the enum keys.

Just a random thought:  I think it would make sense for a user to both be an admin and a member (like a group lead student maybe).  If this is the case, map with userID keys is will no longer work (one of the solutions we were discussing yesterday).


Also, I would highly suggest not storing a list of taskIDs here.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
godsfavours: Maintain use of enum

godsfavours:
> > Admin and Member users, along with any new roles in the future, now stored in HashMap<UserProjectRole, HashSet> in java. HashSets of each member role are stored independently in datastore (admins have their own hashset, etc.). Thoughts on this approach?
> 
> Storing as multiple flat lists in datastore I think would be a good idea (as long as we don't have plans to add more user roles later on). It definitely would make filtering by these IDs easier. If we're doing it this way, we probably don't need to build the complex hashmap in java as well. If you keep them as separate sets, it'll more closely match the data model, but you won't have to deal with the enum keys.
> 
> Just a random thought: I think it would make sense for a user to both be an admin and a member (like a group lead student maybe). If this is the case, map with userID keys is will no longer work (one of the solutions we were discussing yesterday).
> 
> Also, I would highly suggest not storing a list of taskIDs here.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -41,32 +46,51 @@ public ProjectData(String name, String description, long creatorId) {
...
+    parseEntityHashSets(entity, PROPERTY_MEMBERS);
+  }
+
+  /**
+   * Used by ProjectData(Entity) to parse entity properties into respective HashSets of this class
+   *
+   * @param entity the entity
+   * @param propertyName the name of entity property
+   */
+  private void parseEntityHashSets(Entity entity, String propertyName) {
```
jcespejo, line 68: Is this still needed?  I believe you can cast directly from a datastore array to a java arraylist

PR #52, opened 2020-08-07
Add Methods to TaskController

Tzirliya:
More methods are added to TaskController, along with corresponding unit tests. Recursive functions are wrapped in a transaction.

`getTasksByUserID()` will be used by a user servlet to fetch all tasks that a user is a member of.
`getTasksByProjectID()` will be used by a project servlet to fetch all of the top-level tasks under a project.
`addUser()` will be used by a task servlet to recursively add a user to the current task and its parent tasks.
`removeUser()` will be used by a task servlet to recursively remove a user from the current task and its subtasks.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
@@ -50,10 +50,13 @@ public final class TaskController {
   }
 
   public void addUser(TaskData task, long userID) {
-    if (!task.getUsers().contains(userID)) {
-      task.getUsers().add(userID);
-      if (task.getTaskID() != 0) {
+    if (task.getTaskID() != 0) {
+      if (!task.getUsers().contains(userID)) {
+        task.getUsers().add(userID);
         datastore.put(task.toEntity());
+        if (task.getParentTaskID() != 0) {
+          addUser(task.getParentTaskID(), userID);
+        }
       }
     }
   }
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
@@ -50,28 +50,10 @@ public final class TaskController {
   }
 
   public void addUser(TaskData task, long userID) {
-    if (task.getTaskID() != 0) {
-      TransactionOptions options = TransactionOptions.Builder.withXG(true);
-      Transaction transaction = datastore.beginTransaction(options);
-      try {
-        addUserRecursive(task, userID);
-        transaction.commit();
-      } finally {
-        if (transaction.isActive()) {
-          transaction.rollback();
-        }
-      }
-    }
-  }
-
-  private void addUserRecursive(TaskData task, long userID) {
-    if (task.getTaskID() != 0) {
-      if (!task.getUsers().contains(userID)) {
-        task.getUsers().add(userID);
+    if (!task.getUsers().contains(userID)) {
+      task.getUsers().add(userID);
+      if (task.getTaskID() != 0) {
         datastore.put(task.toEntity());
-        if (task.getParentTaskID() != 0) {
-          addUserRecursive(getTaskByID(task.getParentTaskID()), userID);
-        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -54,6 +126,21 @@ public TaskData getTaskByID(long taskID) {
...
+    return getTasks(filter, NO_QUERY_LIMIT, NO_QUERY_SORT);
+  }
+
+  public ArrayList<TaskData> getTasksByProjectID(long projectID) {
+    Filter filter =
+        new CompositeFilter(
+            CompositeFilterOperator.AND,
+            Arrays.asList(
+                FilterOperator.EQUAL.of("projectID", projectID),
+                FilterOperator.EQUAL.of("parentTaskID", 0)));
```
jcespejo, line 140: This wouldn't return subtasks I believe?  This should be ok, just make sure to document somewhere (comment here or design doc) and clarify with the rest of the team that the project page will only display the first level tasks.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -42,6 +43,77 @@ private void addKeysToTasks(ArrayList<TaskData> tasks, ArrayList<Key> keys) {
...
+    }
+  }
+
+  private void addUserRecursive(TaskData task, long userID) {
+    if (task.getTaskID() != 0) {
+      if (!task.getUsers().contains(userID)) {
+        task.getUsers().add(userID);
+        datastore.put(task.toEntity());
+        if (task.getParentTaskID() != 0) {
+          addUser(task.getParentTaskID(), userID);
```
jcespejo, line 73: This would add a user on a subtask to all its ancestors.  This is an interesting concept.  Can you describe some use cases?

I'm thinking for example if I have a task called "Facilitate Capstone project", then I create a subtask called "Build Capstone features" which I assign to you, IMO you shouldn't need to be on my parent "Facilitate" task.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -42,6 +43,77 @@ private void addKeysToTasks(ArrayList<TaskData> tasks, ArrayList<Key> keys) {
...
+        transaction.commit();
+      } finally {
+        if (transaction.isActive()) {
+          transaction.rollback();
+        }
+      }
+    }
+  }
+
+  private void addUserRecursive(TaskData task, long userID) {
```
jcespejo, line 67: Maybe ok for now, but for bigger projects definitely avoid doing recursive database operations.  Transactions does protect us from failures in between.  But we'd still be doing multiple back-and-forths with the database server which may take some time.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
Tzirliya: Yeah I thought returning all subtask layers would be more complicated and wouldn't be useful for most scenarios where someone would just want to see the general idea of a project. I was also thinking that maybe after mvp, an extra feature can be added where a user can choose the amount of layers of nested tasks to display. I can add this to the design doc.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
Tzirliya: I guess that depends on the definition of being a user assigned to a task. I figured that it's more like whether a user is working on that task (since subtask generally implies being a feature of the parent task, working on a subtask would mean working on the parent task). It seems like your definition is whether a user is responsible for completing that task.

I can change is so that the `addUser()` and `removeUser()` methods are not recursive, but I'd like to hear more opinions first so that I'm not undoing and redoing everything.

jcespejo: Good points.  Probably either document on the design doc and ask for comments, or post on chat.

One other thing to consider, who's allowed to mark a status as Completed?  If we have subtask owners listed on the parent task as a user, will we need the concept of task-owners or task-admins?

PR #53, opened 2020-08-10
Add/Remove Users from Tasks

Tzirliya:
Servlets are added to add and remove users from tasks. Buttons are created on the task page to add or remove the users. The add task feature is updated to add parentTaskID when creating subtasks.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
@@ -79,6 +79,7 @@ public class TaskServlet extends HttpServlet {
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get and create parameters
+    long parentTaskID = Long.parseLong(request.getParameter("parentTaskID"));
     long projectID = Long.parseLong(request.getParameter("projectID"));
     String name = request.getParameter("name").trim();
     String description = request.getParameter("description").trim();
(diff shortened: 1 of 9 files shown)


src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
```diff
@@ -41,7 +41,9 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     // Get Task
     long taskID = Long.parseLong(request.getParameter("taskID"));
     TaskController taskController = new TaskController(datastore);
-    // TaskData task = taskController.getTaskByID(taskID);
+    if (taskID != 0 && taskID != 1) {
+      task1 = taskController.getTaskByID(taskID);
+    }
     // ArrayList is for HashMap below. Is there a better way to do this?
     ArrayList<TaskData> taskInArrayList = new ArrayList<>(Arrays.asList(task1));
```
godsfavours, line 48: Is it necessary to put this ArrayList in a HashMap? I believe you could just send this ArrayList directly to the response writer in line 76

Tzirliya: I can, but I'm putting it into a HashMap because later I'm going to uncomment most of the code below (lines 50-65) and put all of that in the HashMap as well (lines 72-75) and send that to the response writer all together. Then everything will be clearly accessible on the JavaScript end (lines 19-23 in task.js).

PR #54, opened 2020-08-10
More Methods/Tests for Authentication

talgsternberg:


src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -61,22 +67,29 @@ public void loginUser(HttpServletRequest request, HttpServletResponse response)
...
+    currCookie = generateCurrCookie(request, response);
 
     // get cookie value for user
     String currUserIDString = currCookie.getValue();
     return Long.parseLong(currUserIDString);
   }
+
+  // logs out user by setting cookie value to -1
+  public void logoutUser(HttpServletRequest request, HttpServletResponse response) {
+    currCookie = generateCurrCookie(request, response);
```
jcespejo, line 81: We don't need to check for the current cookie in the request for logout.  We can just blindly always set to NO_LOGGED_IN_USER on the response.  This would be like if  you click Logout while you're logged in, vs. if you click Logout while you're not currently logged in, they would (should) probably behave the same way.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -61,22 +67,29 @@ public void loginUser(HttpServletRequest request, HttpServletResponse response)
     }
   }
 
+  // returns the userID long (or -1) associated with the user logged in
   public long whichUserLoggedIn(HttpServletRequest request, HttpServletResponse response) {
```
jcespejo, line 71: nit:  "whichUser**Is**LoggedIn" or "getLoggedInUser"

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -27,7 +27,8 @@ public AuthOps(DatastoreService ds) {
     UserController controller = new UserController(ds);
   }
 
-  public void loginUser(HttpServletRequest request, HttpServletResponse response) {
+  // generates cookie. Used in most methods.
+  public Cookie generateCurrCookie(HttpServletRequest request, HttpServletResponse response) {
```
jcespejo, line 31: Good reusable method.  Although it doesn't really generate a cookie, but more so it extracts the cookie from the request.  Maybe something like "getCookieFromRequest" would work better.

(Also if the response is not needed, please remove from the params)

src/test/java/com/rtb/projectmanagementtool/auth/AuthOpsTest.java
```diff
@@ -141,7 +141,48 @@ public void testLoginUser() {
   public void testWhichUser() {
     setUserServiceAuthInfo(false, "abc"); // user is logged in
```
jcespejo, line 142: would be "true" if they're logged in I believe?

PR #55, opened 2020-08-10
Project display servlet & other small changes to classes

godsfavours:


src/main/java/com/rtb/projectmanagementtool/project/ProjectDisplayServlet.java
```diff
@@ -0,0 +1,44 @@
...
+            projectController.getProjectById(Long.parseLong(request.getParameter("projectId")));
+        response.getWriter().println(project);
+        break;
+        // Get projects for the home page
+      case "home":
+        ArrayList<ProjectData> userProjects =
+            projectController.getProjectsWithUser(Long.parseLong(request.getParameter("userId")));
+        response.getWriter().println(userProjects);
+        break;
+    }
```
Tzirliya, line 42: Is there a reason for combining these onto one servlet? It looks like it would be simpler to have two separate servlets because they would both need to get different information.

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -236,16 +220,21 @@ public boolean equals(Object other) {
...
+    json += "\"" + this.description + "\",";
+    json += "\"creatorId\":";
+    json += "\"" + this.creatorId + "\",";
+    json += "\"members\":";
+    json += "\"" + this.members.toString() + "\",";
+    json += "\"admins\":";
+    json += "\"" + this.admins.toString() + "\"";
+    json += "}";
+    return json;
   }
```
Tzirliya, line 239: I'm assuming you changed `toString()` to print ProjectData as JSON because that's how it'll be used in the servlet. Is there a reason for doing it like this manually instead of using Gson, like `gson.toJson(your_data)`?

godsfavours: gson.toJson(ProjectData) includes the private variables of ProjectData in the resulting Json, and while it shouldn't affect functionality, it doesn't look good. Depending on how javascript reacts to this I might use gson.toJson

src/main/java/com/rtb/projectmanagementtool/project/ProjectDisplayServlet.java
godsfavours: Yep, I'll most likely separate them

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
Tzirliya: That makes sense.

jcespejo: If that's the case `toString` maybe shouldn't be used to send back to the JS as part of the response.  `toString` by convention is used to dump the whole class onto a string (eg. for debugging purposes) and would probably be ideal to also return private members.

If you need a version that does not include specific members for JS consumption, consider adding a "formatForResponse()" method or something similar

src/test/java/com/rtb/projectmanagementtool/project/ProjectDisplayServletTest.java
```diff
@@ -0,0 +1,97 @@
...
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter printWriter = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(printWriter);
+    when(request.getParameter("page")).thenReturn("project");
+    when(request.getParameter("projectId")).thenReturn("3"); // hard-coded
+
+    // Run Servlet
+    servlet.doGet(request, response);
+
+    String expectedOutput = newProject3.toString().replaceAll(" ", "");
```
jcespejo, line 72: Probably not a good idea to assert against strings if you need to reformat one of them.  Also think about the hassle of changing field orders and having to redo both sides.  Maybe look into parsing both in json and validating against the json object.

PR #56, opened 2020-08-10
Add Methods to Update Task's Status

Tzirliya:
Methods are added to TaskController to set the status of a task as Complete or Incomplete. Users can only set a task as Complete if all of the tasks subtasks are set to Complete. Users can only set a task as Incomplete if the task's parent task is not set as Complete. Unit tests are added.


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -71,6 +71,38 @@ public void removeUser(TaskData task, long userID) {
...
   }
 
+  public void setComplete(long taskID) {
+    setComplete(getTaskByID(taskID));
+  }
+
+  public void setComplete(TaskData task) {
+    ArrayList<TaskData> subtasks = getSubtasks(task);
+    for (TaskData subtask : subtasks) {
+      if (subtask.getStatus() != Status.COMPLETE) {
```
jcespejo, line 81: This validates if all subtasks are completed before setting the complete flag?  I think this is a good idea.  I would suggest splitting this up into two methods though.  There could be cases where you'd want to just validate the subtasks without actually changing the parent  task (maybe for changing a UI element or something).  There could also be scenarios where you'd need to change the parent task status blindly without looking at the subtasks first.

ie:
```
if (!allSubtasksCompleted(task) {
   return
}

setComplete(task);
```

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -71,6 +71,38 @@ public void removeUser(TaskData task, long userID) {
...
+  }
+
+  public void setIncomplete(long taskID) {
+    setIncomplete(getTaskByID(taskID));
+  }
+
+  public void setIncomplete(TaskData task) {
+    Long parentTaskID = task.getParentTaskID();
+    if (parentTaskID != 0 && getTaskByID(parentTaskID).getStatus() == Status.COMPLETE) {
+      return;
```
jcespejo, line 98: If there's a failure scenario, try to see if you can have your method return the status.  That way callers can find out if the set method was successful or not.

src/main/java/com/rtb/projectmanagementtool/task/TaskSetStatusServlet.java
```diff
@@ -0,0 +1,39 @@
...
+    Status status = Status.valueOf(request.getParameter("status").trim().toUpperCase());
+
+    // Set status
+    TaskController taskController = new TaskController(datastore);
+    if (status == Status.COMPLETE) {
+      taskController.setComplete(taskID);
+    } else if (status == Status.INCOMPLETE) {
+      taskController.setIncomplete(taskID);
+    }
+  }
```
jcespejo, line 38: What's the plan after calling this?  Would the page update the display?  How would the page know if changing the status was successful or not?

Tzirliya: I'm planning on reloading the task page info via the javascript function `getTaskInfo()` that's usually called in body onload. Maybe there's a more efficient way to only change the updated info, but for now I'm just going with the simplest option of fetching all of the task's info on each update.

jcespejo: sgtm to go with simple approach first.

For partial loading, you can have the servlet return a json status code, then update the html page during fetch().then() when the response is received.  
eg. `if response.error=0 {document.getElementById('THE_TASK').setClassName('completed');}`

PR #58, opened 2020-08-11
Servlet to get data for home page

godsfavours:


src/main/java/com/rtb/projectmanagementtool/pages/HomePageServlet.java
```diff
@@ -0,0 +1,91 @@
+/** Servlet responsible getting data for loading home page */
+package com.rtb.projectmanagementtool.pages;
```
jcespejo, line 2: We started off by packaging them by feature (user/project/task/auth, etc..), but now it's looking like we're packing them by layer.  If possible, I would suggest keeping your packaging convention consistent.

src/main/java/com/rtb/projectmanagementtool/pages/HomePageServlet.java
```diff
@@ -0,0 +1,91 @@
...
+
+    // Get user's tasks
+    ArrayList<TaskData> userTasks = taskController.getTasksByUserID(userId);
+
+    response
+        .getWriter()
+        .println(getJsonForResponse(/*userLoggedIn*/ true, user, userProjects, userTasks));
+  }
+
+  public String getJsonForResponse(
```
jcespejo, line 63: This is a good idea to allow for returning different components in the response.  I would suggest moving this into a "utils" method so all the other servlets can reuse.

It would also be nice if the javascript has a reusable response parser that would take this big chunk, split it into the different components, then delegate into more specific methods to process.

@Tzirliya @talgsternberg heads up, for consistenty

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -10,13 +10,6 @@
 import java.util.HashSet;
 
 public class ProjectData {
-  private final String PROPERTY_NAME = "name";
-  private final String PROPERTY_CREATOR = "creator";
-  private final String PROPERTY_DESCRIPTION = "description";
-  private final String PROPERTY_TASKS = "tasks";
-  private final String PROPERTY_ADMINS = "admins";
-  private final String PROPERTY_MEMBERS = "members";
```
jcespejo, line 18: I thought the constants were a good idea.  Curious why you switched to hard coded property names

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
```diff
@@ -220,21 +212,16 @@ public boolean equals(Object other) {
...
-    json += "\"" + this.description + "\",";
-    json += "\"creatorId\":";
-    json += "\"" + this.creatorId + "\",";
-    json += "\"members\":";
-    json += "\"" + this.members.toString() + "\",";
-    json += "\"admins\":";
-    json += "\"" + this.admins.toString() + "\"";
-    json += "}";
-    return json;
+    String returnString = "{\n";
```
jcespejo, line 218: nit:  using curly braces to wrap around stuff, it would be nice if the stuff inside were indented.

src/main/java/com/rtb/projectmanagementtool/task/Status.java
```diff
@@ -0,0 +1,7 @@
+package com.rtb.projectmanagementtool.task;
+
+/** Enum containing status options for a task. */
+public enum Status {
+  COMPLETE,
+  INCOMPLETE
+}
```
jcespejo, line 7: Instead of creating a separate file for this, would this have worked as a public member of TaskData?

```
class TaskData {
  public enum Status {
    ...
  }
}
```

src/main/java/com/rtb/projectmanagementtool/pages/HomePageServlet.java
```diff
@@ -0,0 +1,91 @@
...
+      response
+          .getWriter()
+          .println(
+              getJsonForResponse(
+                  /*userLoggedIn*/ false,
+                  /*user*/ null,
+                  /*userProjects*/ null,
+                  /*userTasks*/ null));
+      return;
+    }
```
jcespejo, line 48: This is a good starting point for the authentication piece.  I would suggest moving this up higher though, so we don't have to initialize the controllers if the user is not logged in anyways. Then maybe adding a "BuildNotLoggedInResponse()" method (which passes the false/null/null/null params to the response builder) just so we don't have to pass the false/null params every time.

@Tzirliya @talgsternberg

src/main/java/com/rtb/projectmanagementtool/project/ProjectData.java
godsfavours: The constants would appear in the json that was created in the HomePageServlet. It technically doesn't affect functionality but it looked pretty out of place

jcespejo: I see.  I did a quick search, looks like tagging them with the "transient" keyword would prevent serialization into json.  I'm not sure if tagging them as "static" would as well

src/main/java/com/rtb/projectmanagementtool/task/Status.java
godsfavours: I tried, but it only works when it's in its own file

PR #59, opened 2020-08-11
Create New User

talgsternberg:


src/main/webapp/user_settings.html
```diff
@@ -36,7 +36,7 @@ <h3>Change Skills:</h3>
           </ul>
         </form>
         <h3>Change Other Information:</h3>
-        <form id="stats-form">
+        <form id="stats-form" action="/create-user" method="POST">
```
jcespejo, line 39: Try to build all of these in a single form.  Each form would be 1 submit to the servlet, and you'd want the submit to have complete information to be able to create the user

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
```diff
@@ -1,9 +1,8 @@
-package com.rtb.projectmanagementtool.auth;
```
jcespejo, line 1: I would suggest keeping this in its own auth package

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  // for testing
+  public CreateUserServlet(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // call auth service to generate AuthID
```
jcespejo, line 30: Move into authOps as its own method.  Ideally we'd want all other pieces to not know about UserService (aka encapsulation).  This way if later on we decide to scrap UserService and replace with something else, you'd only need to touch that one file.

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -0,0 +1,66 @@
...
+    long userID = (long) entity.getKey().getId();
+
+    // set AuthID in entity
+    entity.setProperty("AuthID", AuthID);
+
+    // new AuthOps object
+    AuthOps auth = new AuthOps(datastore);
+
+    // get/create cookie and set value to userID
+    cookie = auth.getCurrCookie(request);
```
jcespejo, line 56: Also move into AuthOps, eg `setLoggedInUser(userID)`.  This way we isolate the whole cookie handling in AuthOps only

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -0,0 +1,66 @@
...
+    String userIDString = Long.toString(userID);
+    cookie.setValue(userIDString);
+
+    // redirect to user_settings
+    response.sendRedirect("/user_settings.html");
+
+    // now how do I update the rest of my user's fields given their input
```
jcespejo, line 63: You can set them the same time you do the initial build/insert (line 38)

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -0,0 +1,66 @@
...
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // call auth service to generate AuthID
+    UserService userService = UserServiceFactory.getUserService();
+    String AuthID = userService.getCurrentUser().getUserId();
+
+    // initialize userMajors
+    ArrayList<String> majors = new ArrayList<>();
+
+    // create new UserData object with UserID
+    UserData newUser = new UserData("", "", 0, majors, Skills.NONE, 0);
```
jcespejo, line 38: Try to move these into your controller.  The servlet (view) should only know about building the view.  The parts that interacts with the data (datastore) shouldn't be visible here.

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
Tzirliya, line 63: Maybe create servlets for each form in the user_settings.html page. When a user clicks "submit" in that form, it'll set a POST request to that servlet. Then, in the servlet, use `request.getParameter("parameter_name")` to get all of the user's input.

Alternatively, if you want to keep it all in one servlet, you can add a hidden input to the html that tells you which form was submitted. (Maybe there's a better way to get the form id without using a hidden input?) Then check in the servlet which form was submitted and make updates accordingly. I'd recommend making some UserController methods to update different attributes so you don't have to write all of the logic here.

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // to entity
+    Entity entity = newUser.toEntity();
+
+    // put in datastore
+    datastore.put(entity);
```
Tzirliya, line 44: Why don't you use your UserController to addUser()?

src/main/java/com/rtb/projectmanagementtool/user/AuthOps.java
talgsternberg: Only issue is I'll have trouble accessing/changing Skills later...

jcespejo: See @godsfavours's PR:  https://github.com/googleinterns/step171-2020/pull/58

He's trying to setup enums so it'll be accessible publicly.  I had just sent him this example:  https://stackoverflow.com/questions/6280937/java-how-to-access-inner-enum-class

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -27,8 +27,22 @@ public AuthOps(DatastoreService ds) {
     UserController controller = new UserController(ds);
   }
 
-  // generates cookie. Used in most methods.
-  public Cookie generateCurrCookie(HttpServletRequest request, HttpServletResponse response) {
+  public String getAuthID() {
+    // call auth service to generate AuthID
+    UserService userService = UserServiceFactory.getUserService();
+    String AuthID = userService.getCurrentUser().getUserId();
```
jcespejo, line 33: There wouldn't always be a current user (eg. if user is not logged in), so this might error out with a nullpointerexception.  getAuthID then maybe can return a "not-logged-in" state for when the user is not logged in

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -27,8 +27,22 @@ public AuthOps(DatastoreService ds) {
...
+    return AuthID;
+  }
+
+  // get/create cookie and set value to userID
+  public void createAndSetCookieNewUser(HttpServletRequest request, long userID) {
```
jcespejo, line 38: I'm not sure I understand this method, since Cookie here doesn't go anywhere.  If you intended to send this cookie back to the user's browser, you'll then need to do response.addCookie

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -0,0 +1,60 @@
...
+
+  // for testing
+  public CreateUserServlet(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // new AuthOps object
+    AuthOps auth = new AuthOps(datastore);
```
jcespejo, line 34: Good steps so far.  Probably doable at this point to also get the form data from the parameters and set them into t he blank UserData before pushing into datastore

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -0,0 +1,96 @@
...
+     * userController.getUserByID(userID);
+     *
+     * <p>// new TaskController TaskController taskController = new TaskController(datastore);
+     *
+     * <p>// get this user's tasks ArrayList<TaskData> tasks =
+     * taskController.getTasksByUserID(userID);
+     */
+
+    // Set attributes of request; retrieve in jsp with
+    request.setAttribute("UserProfile", user);
```
jcespejo, line 90: nit:  "UserData"

src/main/webapp/user_profile.jsp
```diff
@@ -0,0 +1,45 @@
+<%--Class Imports--%>
```
jcespejo, line 1: from a quick glance, jsp looks good

PR #64, opened 2020-08-13
Servlets and jsp

talgsternberg:
First attempt at user-settings.jsp and doGet() for UserSettingsServlet to set up page


src/main/java/com/rtb/projectmanagementtool/user/UserSettingsServlet.java
```diff
@@ -0,0 +1,174 @@
...
+      checkedStatus.set(2, "checked");
+    } else if (skillsAsString.equals("WRITING")) {
+      checkedStatus.set(3, "checked");
+    } else if (skillsAsString.equals("ART")) {
+      checkedStatus.set(4, "checked");
+    } else if (skillsAsString.equals("WEBDEV")) {
+      checkedStatus.set(5, "checked");
+    } else {
+      checkedStatus.set(6, "checked");
+    }
```
Tzirliya, line 105: Can a switch-case be used?

src/main/webapp/user_profile.jsp
```diff
@@ -28,16 +28,15 @@
...
           <p>Total Completed Tasks: <%=user.getUserTotal()%></p>
       </div>
       
       <div id="user-tasks-container">
         <h2><%=user.getUserName()%>'s Tasks:</h2>
         <%for (TaskData task : userTasks) {%>
-          <p>Task Name: <%=task.getName()%></p>
-          <p>Task Status: <%=task.getStatus()%></p>
+          <p><%=task.getName()%>, <%=task.getStatus()%></p>
         <%}%>
```
Tzirliya, line 40: You can the list-tasks.jsp file to display a list of tasks. See lines 55-56 in task.jsp to see how I included it in the task page.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -27,18 +27,25 @@ public AuthOps(DatastoreService ds) {
     this.controller = new UserController(ds);
   }
 
-  public String getAuthID() {
```
jcespejo, line 30: for `getAuthID` specifically, imo the previous version was more appropriate.  the concept of an AuthID is separate from the concept of a login cookie.  If you need to check for both, I would advice creating two separate methods (you already have one for checking the login cookie) and calling both one after the other

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -27,18 +27,25 @@ public AuthOps(DatastoreService ds) {
...
+    }
+    // if no one logged in, return empty string
+    return "";
   }
 
   // get/create cookie and set value to userID
-  public void createAndSetCookieNewUser(HttpServletRequest request, long userID) {
+  public void createAndSetCookieNewUser(
+      HttpServletRequest request, HttpServletResponse response, long userID) {
     Cookie cookie = getCurrCookie(request);
```
jcespejo, line 44: If you're overwriting the cookie anyways, no need to search for the old one and replace.  You can just build a new cookie (by filling in name and value) then dump it into the response right away.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -27,18 +27,25 @@ public AuthOps(DatastoreService ds) {
...
+      UserService userService = UserServiceFactory.getUserService();
+      String AuthID = userService.getCurrentUser().getUserId();
+      return AuthID;
+    }
+    // if no one logged in, return empty string
+    return "";
   }
 
   // get/create cookie and set value to userID
-  public void createAndSetCookieNewUser(HttpServletRequest request, long userID) {
```
jcespejo, line 38: nit:  this method would be useful for other use cases as well, and not just for the new user use case.  It would be ok to rename this to something more generic like "setLoggedInCookie"

src/main/java/com/rtb/projectmanagementtool/user/UserSettingsServlet.java
talgsternberg: I'll look into it. Good suggestion, thanks!

PR #66, opened 2020-08-13
Create InitTestData class & updates to servlets and jsp

godsfavours:
InitTestData class creates ProjectData, UserData, and TaskData objects to test the MVP
I've added some ProjectData and UserData objects. Need objects for TaskData.
Feel free to add some more.

To create new UserData objects:
Run appengine
Click log in button
Enter the name you want to use for user in "Email" field
Look in console for line with "AuthID: [id]"
Copy [id]
In InitTestData, when creating the user:
paste [id] into the authId field

Some times when trying to logging in this error message comes up:
```
Error 501 Method ET is not defined in RFC 2068 and is not supported by the Servlet API
```
Not sure what's causing it, but refreshing the page should fix it. A HTTP 500 error might appear after the refresh but another refresh should fix


src/main/java/com/rtb/projectmanagementtool/home/HomeServlet.java
```diff
@@ -16,35 +21,38 @@
...
-    // Mock authentication
-    if (!request.getParameterMap().containsKey("user")) {
-      // Redirect to /login servlet if authentication fails
-      request.getRequestDispatcher("/login").forward(request, response);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Authenticate
+    AuthOps auth = new AuthOps(datastore);
+    auth.loginUser(request, response);
+    Long userLoggedInId = auth.whichUserIsLoggedIn(request, response);
```
jcespejo, line 29: Looks good.  Might be a good idea to encapsulate this in AuthOps as well, maybe something like `getLoggedInUserOrRedirect()`, but I'm not sure how to cleanly handle the early exit yet.

src/main/java/com/rtb/projectmanagementtool/init/InitTestData.java
```diff
@@ -0,0 +1,220 @@
...
+  public void removeAllEntities(DatastoreService datastore) {
+    removeAllEntities(datastore, new Query("User"));
+    removeAllEntities(datastore, new Query("Project"));
+    removeAllEntities(datastore, new Query("Task"));
+  }
+
+  public void removeAllEntities(DatastoreService datastore, Query query) {
+    PreparedQuery results = datastore.prepare(query);
+    for (Entity entity : results.asIterable()) {
+      datastore.delete(entity.getKey());
```
jcespejo, line 71: You can batch delete by passing an array of keys I thought, instead of doing 1 delete per record.

I'm curious now if you can have a single ArrayList of keys combining the keys for users, projects, and tasks (since technically they're all the same object type) and pass that to delete.

src/main/webapp/index.jsp
```diff
@@ -1,2 +1,2 @@
-<!-- Forward to /home servlet upon entering website -->
```
jcespejo, line 1: You can also check for the domain to auto redirect to the correct page.  Prod would be "rtb-step-2020.uc.r.appspot.com".

PR #67, opened 2020-08-14
Add Task Buttons and Update Methods

Tzirliya:
When tasks are set to complete or incomplete, it updates all of its users quantity of completed tasks. There's a button on the Task Page where a user can toggle the task's status between complete and incomplete. There's a button at the bottom of the Task Page to recursively delete that task and its subtasks. Users can only be added or removed when a task is set to complete (to avoid users adding themselves to completed tasks just to increase their quantity of completed tasks stat).


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -95,9 +96,22 @@ private boolean allSubtasksAreComplete(TaskData task) {
...
+      task.setStatus(Status.COMPLETE);
+      if (task.getTaskID() != 0) {
+        datastore.put(task.toEntity());
+        UserController userController = new UserController(datastore);
+        UserData user;
+        for (long userID : task.getUsers()) {
+          try {
+            user = userController.getUserByID(userID);
+            user.setUserTotal(user.getUserTotal() + 1);
+            datastore.put(user.toEntity());
```
jcespejo: I mentioned in one of Godfavour's PR, since `put` can accept a list of Entities, I wonder if we can pass it a single ArrayList containing both the Task and the Users that needs to be updated, so we only have to do 1 db call

src/main/java/com/rtb/projectmanagementtool/task/TaskDeleteServlet.java
```diff
@@ -0,0 +1,42 @@
...
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get parameters
+    long taskID = Long.parseLong(request.getParameter("taskID"));
+
+    // Set status
+    TaskController taskController = new TaskController(datastore);
+    if (taskID != 0 && taskID != 1) {
+      long parentTaskID = taskController.getTaskByID(taskID).getParentTaskID();
+      taskController.deleteTasks(new ArrayList<>(Arrays.asList(taskID)));
```
jcespejo, line 36: How do we let the caller know if this delete fails?

src/main/webapp/task.jsp
```diff
@@ -32,7 +34,25 @@
...
       <div id="task-title-container"><h1><%=task.getName()%></h1></div>
       <div id="task-description-container"><%=task.getDescription()%></div>
-      <div id="task-status-container">Status: <%=task.getStatus()%></div>
+      <div id="task-status-container">
+        <p>Status: <%=task.getStatus()%></p>
+        <%
+            String status;
+            String statusButtonText;
+            if (task.getStatus() != Status.COMPLETE) {
+                status = "COMPLETE";
```
jcespejo, line 43: `Status.COMPLETE.name()` I think would work

src/main/webapp/task.jsp
```diff
@@ -69,22 +95,24 @@
...
-            String servletPage;
-            String buttonText;
-            if (!task.getUsers().contains(userID)) {
-                servletPage = "/task-add-user";
-                buttonText = "Assign me to this task";
-            } else {
-                servletPage = "/task-remove-user";
-                buttonText = "Remove me from this task";
-            }
+            if (task.getStatus() != Status.COMPLETE) {
```
jcespejo, line 98: Is this missing the closing bracket?

Tzirliya: It's on line 115.

PR #68, opened 2020-08-14
CreateUserServlet

talgsternberg:


src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -24,25 +23,58 @@ public CreateUserServlet() {
...
-    // create new "blank" UserData object to put in datastore
+    // create new "blank" UserData object
     UserData newUser = controller.createNewUser();
 
-    // set this user's AuthID
+    // new UserService
+    UserService userService = UserServiceFactory.getUserService();
+
+    // if not logged in, redirect to login link
+    if (!userService.isUserLoggedIn()) {
```
talgsternberg, line 45: if the user is logged out, this should hypothetically redirect them to login

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -24,25 +23,58 @@ public CreateUserServlet() {
...
+      // Get login URL
+      String urlToRedirectToAfterUserLogsIn = "/create-new-user";
+      String loginUrl = userService.createLoginURL(urlToRedirectToAfterUserLogsIn);
+      // redirect
+      response.sendRedirect(loginUrl);
+    }
+    // if logged in, redirect back to create-new-user
+    else if (userService.isUserLoggedIn()) {
```
talgsternberg, line 53: once the user is logged in, this should redirect back to the form

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -24,25 +23,58 @@ public CreateUserServlet() {
...
+      response.sendRedirect("/create-new-user.jsp");
+    }
+
+    // get AuthID
+    String AuthID = auth.getAuthID();
```
talgsternberg, line 58: This is the line that has a null pointer.

Tzirliya: Does this only happen when a user is not logged in?

talgsternberg: Theoretically the user should be logged in when they reach this line. Line 45 begins the process of login by checking they are not logged in and then redirecting them to a generated loginUrl

jcespejo: on your redirects above, you'll probably to do an early return to prevent the rest of the code here from being evaluated.  I'm not sure how instant the response.sedRedirect would be.

src/main/webapp/login.jsp
```diff
@@ -13,6 +13,9 @@
       <a href="${loginUrl}">
         <button>Log In</button>
         </a>
+      <a href="create-new-user.jsp">
```
jcespejo, line 16: you'd want to send them to the servlet instead of to the jsp, since the servlet has the logic for authentication checks

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
Tzirliya: +1 I noticed that at one point my code had two redirects and the second redirect would run. So it seems like the code continues running even if there's a redirect.

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -86,4 +72,18 @@ else if (userService.isUserLoggedIn()) {
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get login URL
+    String urlToRedirectToAfterUserLogsIn = "/create-new-user";
+    String loginUrlNewUser = userService.createLoginURL(urlToRedirectToAfterUserLogsIn);
+
+    // send to request
+    request.setAttribute("loginUrlNewUser", loginUrlNewUser);
+    request.getRequestDispatcher("login.jsp").forward(request, response);
```
talgsternberg, line 84: ERROR: CreateUserServlet.java:[84,54] unreported exception javax.servlet.ServletExcept
ion; must be caught or declared to be thrown

Tzirliya: Add ServletException to line 77 and import it.

src/main/webapp/login.jsp
```diff
@@ -13,6 +13,9 @@
       <a href="${loginUrl}">
         <button>Log In</button>
         </a>
+      <a href="${loginUrlNewUser}">
```
talgsternberg, line 16: This link doesn't direct to anything (even though in CreateNewUser it's defined as a login link that redirects to create-new-user.jsp, which is a form)

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -51,11 +69,18 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  }
 
-    // get the form data from the parameters and set
-    // them into the blank UserData before pushing into datastore
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+    // Get login URL
+    request.setAttribute(
+        "loginUrlNewUser", auth.getLoginLink(/*Return URL*/ "/create-new-user.jsp"));
```
talgsternberg, line 81: creates url that should login and redirect to the form in create-new-user.jsp

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -51,11 +69,18 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
+    // Forward to login page
+    request.getRequestDispatcher("login.jsp").forward(request, response);
```
talgsternberg, line 84: sends url to login.jsp

src/main/java/com/rtb/projectmanagementtool/user/CreateUserServlet.java
```diff
@@ -51,11 +69,18 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    response.sendRedirect("/user_profile.jsp");
+  }
 
-    // get the form data from the parameters and set
-    // them into the blank UserData before pushing into datastore
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+    // Get login URL
+    request.setAttribute(
```
jcespejo, line 80: This code is correct, just in the wrong spot.  The login page you see with the two buttons is created by LoginServlet, so you'd want to add info for this button in that servlet:
https://github.com/googleinterns/step171-2020/blob/master/src/main/java/com/rtb/projectmanagementtool/auth/LoginServlet.java


Then I guess the next question would be what you'll need to do in doGet for CreateUserServlet.  (Optional) You can validate if they're logged in via UserService and do an auto redirect if they're not, in case they manually type in the "site.com/create-user-url" instead of going through the button.  Besides that I think you'll just need to redirect to create-user.jsp to show the form

src/main/java/com/rtb/projectmanagementtool/auth/LoginServlet.java
```diff
@@ -28,9 +28,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+        "loginUrlNewUser", auth.getLoginLink(/*Return URL*/ "/create-new-user.jsp"));
 
     // Forward to login page
     request.getRequestDispatcher("login.jsp").forward(request, response);
+
+    // login the user with cookies
+    auth.loginUser(request, response);
+
+    // redirect back home
+    response.sendRedirect("/home"); // error here
```
talgsternberg, line 44: "Response has already been committed" error. Shouldn't this not be the case bc of the return in the IF?

src/main/java/com/rtb/projectmanagementtool/auth/LoginServlet.java
```diff
@@ -28,9 +28,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     }
 
     // Get login URL
-    request.setAttribute("loginUrl", auth.getLoginLink(/*Return URL*/ "/home"));
+    request.setAttribute("loginUrl", auth.getLoginLink(/*Return URL*/ "/LoginServlet"));
```
jcespejo, line 31: This would be /login I believe

src/main/java/com/rtb/projectmanagementtool/auth/LoginServlet.java
```diff
@@ -28,9 +28,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+
+    // Get login URL for first time users (on submit returns to LoginServlet)
+    request.setAttribute(
+        "loginUrlNewUser", auth.getLoginLink(/*Return URL*/ "/create-new-user.jsp"));
 
     // Forward to login page
     request.getRequestDispatcher("login.jsp").forward(request, response);
+
+    // login the user with cookies
+    auth.loginUser(request, response);
```
jcespejo, line 41: This needs to be moved up to, right before checking for whichUserIsLoggedIn

src/main/java/com/rtb/projectmanagementtool/auth/LoginServlet.java
jcespejo: don't need this redirect anymore since we're already doing this up top.  The error is saying `getRequestDispatcher` already commited the response

PR #69, opened 2020-08-14
Create CommentData and CommentController Classes

Tzirliya:
Suggestions from the previous pull request are implemented. CommentData.java and CommentController.java are created along with unit tests.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/comment/CommentController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/comment/CommentController.java
@@ -1,0 +1,120 @@
+package com.rtb.projectmanagementtool.comment;
+
+import com.google.appengine.api.datastore.*;
+import com.google.appengine.api.datastore.Query.*;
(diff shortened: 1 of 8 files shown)


src/main/java/com/rtb/projectmanagementtool/comment/CommentController.java
```diff
@@ -0,0 +1,120 @@
...
+    if (comments.size() == keys.size()) {
+      for (int i = 0; i < keys.size(); i++) {
+        comments.get(i).setCommentID(keys.get(i).getId());
+      }
+    }
+  }
+
+  // Get methods
+
+  public CommentData getCommentByID(long commentID) {
```
jcespejo, line 40: ooc what use cases would we return only a single comment?

Tzirliya: I'm not sure yet, but I figure it's useful to have.

PR #70, opened 2020-08-14
Implement suggestions from prev PR

godsfavours:


src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -16,7 +16,7 @@
   // public String cookieValue;
   public Cookie currCookie;
   private final String COOKIENAME = "sessionUserID";
-  private final String NO_LOGGED_IN_USER = "-1"; // -1 = logged out
+  public static final String NO_LOGGED_IN_USER = "-1"; // -1 = logged out
```
jcespejo, line 19: nit:  I would prefer storing this as a long here, and do the string->long casting only in this class.  This way customers of this helper class wouldn't have to worry about casting.

A lot of the spots here that initializes with NO_LOGGED_IN_USER also probably doesn't need to start off with -1 (for example setLoggedInCookie can just set the userID off the bat), so that'll reduce the casts needed as well

src/main/java/com/rtb/projectmanagementtool/init/InitTestData.java
```diff
@@ -33,18 +33,16 @@
 
   // The tasks
 
+  private ArrayList<Key> entityKeys; // arrays to store entity keys for batch delete
```
jcespejo, line 36: probably doesn't need to be a class member since this is a temporary list.   Plus servlets don't hold state anyways across requests

PR #72, opened 2020-08-17
Add and Delete Comments of a Task

Tzirliya:
Users can post comments on a Task Page. Comments are displayed on the Task Page. Users can delete their own comments from the Task Page. Users cannot add or delete comments if a task is set to COMPLETE. Servlets and unit tests are created to do these features.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/comment/CommentDeleteServlet.java
+++ b/src/main/java/com/rtb/projectmanagementtool/comment/CommentDeleteServlet.java
@@ -1,0 +1,40 @@
+package com.rtb.projectmanagementtool.comment;
+
+import com.google.appengine.api.datastore.*;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns task data */
+@WebServlet("/comment-delete")
+public class CommentDeleteServlet extends HttpServlet {
+
+  DatastoreService datastore;
+
+  public CommentDeleteServlet() {
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+
+  // For testing only
+  public CommentDeleteServlet(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get parameters
+    long commentID = Long.parseLong(request.getParameter("commentID"));
+    long taskID = Long.parseLong(request.getParameter("taskID"));
+
+    // Delete comment
+    CommentController commentController = new CommentController(datastore);
+    commentController.deleteComments(new ArrayList<>(Arrays.asList(commentID)));
+
+    // Redirect to comment's task page
+    response.sendRedirect("/task?taskID=" + taskID);
+  }
+}
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
@@ -1,6 +1,7 @@
 package com.rtb.projectmanagementtool.task;
 
 import com.google.appengine.api.datastore.*;
+import com.rtb.projectmanagementtool.comment.*;
 import com.rtb.projectmanagementtool.project.*;
 import com.rtb.projectmanagementtool.user.*;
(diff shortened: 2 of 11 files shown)


src/main/java/com/rtb/projectmanagementtool/comment/CommentDeleteServlet.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  // For testing only
+  public CommentDeleteServlet(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get parameters
```
jcespejo, line 29: Also check credentials here

src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
```diff
@@ -31,6 +34,20 @@ public TaskServlet(DatastoreService datastore) {
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws ServletException, IOException {
 
+    // Authenticate
+    AuthOps auth = new AuthOps(datastore);
+    auth.loginUser(request, response);
```
jcespejo, line 39: Don't need to login here.  Login servlet should handle login.  all other should just do whichUser then redirect if there's no user logged in

src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
```diff
@@ -74,18 +90,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
     }
 
-    // // Get Comments
+    // Get Comments
     // int quantity = Integer.parseInt(request.getParameter("quantity"));
     // String sortBy = request.getParameter("sortBy");
     // String sortDirection = request.getParameter("sortDirection");
+    CommentController commentController = new CommentController(datastore);
+    ArrayList<CommentData> comments = new ArrayList<>();
+    HashMap<CommentData, String> commentsMap = new HashMap<>();
```
jcespejo, line 99: I would advice creating a new object to hold this aggregation, instead of piggybacking off of the map's key/value pair.  Maybe something like `CommentDisplayData`.  This would be useful in the future as well if you need to add more  derived info in there.

src/main/webapp/navigation-bar.jsp
```diff
@@ -1,7 +1,7 @@
 <div id="navigation">
   <ul>
     <li><a href="/home">Home</a></li>
-    <li><a href="/task?taskID=1">Task Page</a></li>
+    <li><a href="/task?taskID=0">Task Page</a></li>
```
jcespejo, line 4: What's the significance of taskID=0?  If this is just for preventing lookups if a valid taskID was not provided, the servlet should be the one checking for the validity of the taskID and skip the relevant steps if the taskID is not provided

src/main/webapp/task.jsp
```diff
@@ -109,23 +115,60 @@
...
+            Posted on <%=date%> at <%=time%>
+            <br>
+            Posted by <%=username%>
+          </h5>
+          <p><%=comment.getMessage()%></p>
+          <% if (comment.getUserID() == user.getUserID() && task.getStatus() != Status.COMPLETE) {%>
+          <form id="delete-comment-post-form" action="/comment-delete" method="POST">
+            <input type="hidden" id="delete-comment-commentID-input" name="commentID" value="<%=comment.getCommentID()%>">
+            <input type="hidden" id="delete-comment-taskID-input" name="taskID" value="<%=task.getTaskID()%>">
+            <button type="submit"><span class="fa fa-trash-o" aria-hidden="true"></span></button>
```
jcespejo, line 167: this could be a good use case for an ajax delete so you won't have to refresh the whole page after deleting a comment.  `remove()` I think is the function to remove from DOM?  https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove

src/test/java/com/rtb/projectmanagementtool/comment/CommentServletTest.java
```diff
@@ -0,0 +1,71 @@
...
+    int quantityBefore = datastore.prepare(new Query("Comment")).countEntities();
+
+    // Run doPost()
+    servlet.doPost(request, response);
+
+    // Get quantity of tasks after posting
+    int quantityAfter = datastore.prepare(new Query("Comment")).countEntities();
+
+    // Assert a task entity was added to datastore
+    Assert.assertEquals("doPost", quantityAfter, quantityBefore + 1);
```
jcespejo, line 69: also assert that the correct comment and fields were inserted into the datastore

PR #73, opened 2020-08-17
Create servlet to create new projects

godsfavours:


src/main/java/com/rtb/projectmanagementtool/project/CreateProjectServlet.java
```diff
@@ -0,0 +1,55 @@
...
+public class CreateProjectServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Authenticate
+    AuthOps auth = new AuthOps(datastore);
+    auth.loginUser(request, response);
```
jcespejo, line 22: Don't need to login here.  Login servlet should handle logging in.  Just need to check for whichUser here, then redirect if not logged in

src/main/java/com/rtb/projectmanagementtool/project/CreateProjectServlet.java
```diff
@@ -0,0 +1,55 @@
...
+    request.setAttribute("userId", userLoggedInId);
+
+    // Forward to create project page
+    request.getRequestDispatcher("create-project.jsp").forward(request, response);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+
```
jcespejo, line 40: Also check credentials here.  We wouldn't want them creating new projects if they're not logged in.

PR #74, opened 2020-08-18
prechecked buttons on settings form

talgsternberg:


src/main/java/com/rtb/projectmanagementtool/user/UserSettingsServlet.java
```diff
@@ -59,30 +59,30 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     majorsString = majorsString.replaceAll("\\s", "");
 
     // pre checked buttons
-    ArrayList<String> checkedStatus = new ArrayList<>(Arrays.asList("", "", "", "", "", "", ""));
+    String[] checkedStatus = {"", "", "", "", "", "", ""};
     Skills getSkills = user.getUserSkills();
     String skillsAsString = getSkills.name();
 
     // currently not using: look up "switch-case"
     if (skillsAsString.equals("NONE")) {
```
jcespejo, line 67: nit:  map with the skill name as the key would be more readable (and easier to build) since it'll more clearly show which skills are enabled

src/main/webapp/user-settings.jsp
```diff
@@ -40,19 +41,19 @@
         <tr>
           <td>Top Skill:</td><br>
           <td>
-            <input type="radio" value="NONE" name="skills" checked>
+            <input type="radio" value="NONE" name="skills" <%=checkedStatus[0]%>>
             <label>None</label><br>
-            <input type="radio" value="LEADERSHIP" name="skills">
+            <input type="radio" value="LEADERSHIP" name="skills" <%=checkedStatus[1]%>>
```
jcespejo, line 46: As per comment above, clarity can be improved if this was something like `<%=checkedStatus["LEADERSHIP"]%>`

PR #79, opened 2020-08-19
Remove InitTestData class and create servlet for adding users to projects

godsfavours:


src/main/java/com/rtb/projectmanagementtool/user/UserController.java
```diff
@@ -13,6 +13,20 @@ public UserController(DatastoreService datastore) {
     this.datastore = datastore;
   }
 
+  // Only way to find users to add to projects currently
+  // is with their name; means that we'd need to make sure there
+  // are no duplicate names in database
```
Tzirliya, line 18: It doesn't sound like a good idea to make sure all usernames are unique...

src/main/webapp/project-add-user-form.jsp
```diff
@@ -0,0 +1,27 @@
...
+<%
+    Long userId = (Long) request.getAttribute("userId");
+    ProjectData project = (ProjectData) request.getAttribute("project");
+%>
+
+<form action="/add-user-to-project" method="POST">
+  <input type="hidden" name="projectId" value=<%=project.getId()%>>
+  <ul>
+    <li>
+      <input type="text" name="user-name" placeholder="UserName" />
```
Tzirliya, line 12: Maybe use emails instead? They're more likely to be unique than usernames.

src/main/java/com/rtb/projectmanagementtool/project/AddUserToProjectServlet.java
```diff
@@ -0,0 +1,56 @@
...
+      case "admin":
+        project.addAdminUser(user.getUserID());
+        break;
+      case "member":
+        project.addMemberUser(user.getUserID());
+        break;
+    }
+
+    // save project to database
+    projectController.addProject(project);
```
Tzirliya, line 51: nit: you can create ProjectController methods like `addAdmin(project, userID)` and `addMember(project, userID)` or even `addUserToProject(project, userID, role)` which can cover all of it.

src/main/java/com/rtb/projectmanagementtool/auth/AuthOps.java
```diff
@@ -74,8 +74,10 @@ public Cookie getCurrCookie(HttpServletRequest request) {
   }
 
   // logs in user and adds cookie with value of String userID to response
-  public void loginUser(HttpServletRequest request, HttpServletResponse response) {
+  public long loginUser(HttpServletRequest request, HttpServletResponse response) {
     currCookie = getCurrCookie(request);
+    long userId = -1l;
```
jcespejo, line 79: `NO_LOGGED_IN_USER` (constant probably needs to be a long since the majority of the checks against it are for long comparisons)

src/main/java/com/rtb/projectmanagementtool/auth/LoginDispatchServlet.java
```diff
@@ -0,0 +1,43 @@
+// Servlet called after a user logs in with Google. Tt checks to see if
+// their authId matches any entities in datastore, in which case it
+// would direct the user to the home page. If not, it directs them to
+// the page to create a new user.
+// Might need a better name than Login'Dispatch'Servlet
```
jcespejo, line 5: I thought what you have here also would have been a good fit in loginServlet.  It could for example:
1.  Check if user is already logged in, send to home
2.  Check if user has auth, if not send to Google auth and redirect back to loginServlet
3.  If user has auth, find userID and redirect to home
4.  If user has auth but no userID (or if user explicitly clicks on Create User), send to create user


Both servlets as it stand now do similar actions:  check auth situation and redirect somewhere.  Keeping in 1 place I think would make it easier to deal with

src/main/java/com/rtb/projectmanagementtool/project/AddUserToProjectServlet.java
```diff
@@ -0,0 +1,56 @@
...
+    // get request parameters
+    Long projectId = Long.parseLong(request.getParameter("projectId"));
+    String userName = request.getParameter("user-name");
+    String userRole = request.getParameter("user-role");
+
+    // create objects
+    ProjectData project = projectController.getProjectById(projectId);
+    UserData user = userController.getUserByName(userName);
+
+    // if user is not in database or they are already in the project,
```
jcespejo, line 33: Probably should let the user know about this.  We don't want them thinking they added someone but the system actually didn't

src/main/java/com/rtb/projectmanagementtool/user/UserController.java
```diff
@@ -13,6 +13,20 @@ public UserController(DatastoreService datastore) {
     this.datastore = datastore;
   }
 
+  // Only way to find users to add to projects currently
+  // is with their name; means that we'd need to make sure there
```
jcespejo, line 17: This is a dangerous assumption.  In a school setting, chances that at least 2 people have the same name are really high.  Probably would be good to come up with a way to get this to work.

src/main/webapp/project-add-user-form.jsp
```diff
@@ -0,0 +1,27 @@
+<%@ page import="com.rtb.projectmanagementtool.project.*" %>
+
+<%
+    Long userId = (Long) request.getAttribute("userId");
+    ProjectData project = (ProjectData) request.getAttribute("project");
+%>
+
+<form action="/add-user-to-project" method="POST">
```
jcespejo, line 8: Would be cool if the user can do multiple operations (add multiple users + create tasks) without having to reload the page every time.  Think about a professor having to create a project for 30 students, each student having 1 task each.  That'll be 60 page loads at least

src/main/java/com/rtb/projectmanagementtool/project/AddUserToProjectServlet.java
godsfavours: I think it's good to encapsulate the adding users feature to the ProjectData class and not letting ProjectController be aware of  it

PR #80, opened 2020-08-19
Improve user-friendliness on Task Page

Tzirliya:
At the top of the Task Page, there's links to all of its parent tasks and its parent project.
The list of user buttons displays user information better.
The status button is a checkmark that can be toggled and appears next to the task's title. Messages appear when hovering over the checkmark. The list of tasks/subtasks also contain a checkmark icon with messages when hovering, but it cannot be toggled there.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
@@ -6,6 +6,7 @@ import com.rtb.projectmanagementtool.task.TaskData.Status;
 import com.rtb.projectmanagementtool.user.*;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 /** Class controlling the TaskData object. */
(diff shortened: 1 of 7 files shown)


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -207,6 +212,28 @@ public TaskData getTaskByID(long taskID) {
...
+  //             .map(Long::parseLong)
+  //             .collect(Collectors.toList());
+  //     Filter filter = new FilterPredicate("taskID", FilterOperator.IN, parentIDs);
+  //     return getTasks(filter, NO_QUERY_LIMIT, NO_QUERY_SORT);
+  //   }
+
+  public ArrayList<TaskData> getParents(TaskData task) {
+    ArrayList<TaskData> parentTasks = new ArrayList<>();
+    long parentTaskID = task.getParentTaskID();
+    while (parentTaskID != 0) {
```
jcespejo, line 228: Do subtasks have the projectID set?  If so, you might be able to build this relationship by loading all tasks in a project

src/main/java/com/rtb/projectmanagementtool/task/TaskServlet.java
```diff
@@ -54,20 +54,21 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
       taskID = 0l;
     }
 
-    // Get Parent Task
-    TaskData parentTask = null;
-    if (taskID != 0 && task.getParentTaskID() != 0) {
-      parentTask = taskController.getTaskByID(task.getParentTaskID());
-    }
+    // Get parent task stack
+    ArrayList<TaskData> parents = taskController.getParents(task);
```
jcespejo, line 67: nit:  "ancestors" maybe?  technically a task can only have a single parent

src/main/webapp/task.jsp
```diff
@@ -39,43 +41,24 @@
...
+        <h1 class="inline"><%=task.getName()%></h1>
+        <%request.setAttribute("task", task);%>
+        <%request.setAttribute("clickable", true);%>
+        <%request.setAttribute("canSetComplete", canSetComplete);%>
+        <%request.setAttribute("canSetIncomplete", canSetIncomplete);%>
+        <jsp:include page="task-status-checkmark.jsp"/>
       </div>
+      <div id="task-description-container" class="description"><%=task.getDescription()%></div>
 
       <h2>Subtasks</h2>
```
jcespejo, line 63: Could be a cool user experience improvement if they can set multiple subtasks status as well from this page, without refreshing the page per operation.  That would save extra clicks from having to navigate down to the subtasks to mark them as complete.

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
Tzirliya: Yes, I might do that in the future if I have enough time.

src/main/webapp/task.jsp
Tzirliya: I considered that, but that would include many more operations to find out for each subtask if it can be set as complete (all of the subtask's subtasks must be set to complete as well). I might still add that as an option in the future if I have enough time.

PR #81, opened 2020-08-20
Private Comments

talgsternberg:
created private comments class and controller (these comments are to be displayed on user profile next to task buttons)


src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
```diff
@@ -0,0 +1,102 @@
+package com.rtb.projectmanagementtool.privatecomment;
+
+import com.google.appengine.api.datastore.*;
+import com.google.appengine.api.datastore.Query.*;
+import com.rtb.projectmanagementtool.user.*;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Class controlling the CommentData object. */
```
Tzirliya, line 9: nit: old comment

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
```diff
@@ -0,0 +1,102 @@
...
+    PrivateCommentData privateComment = new PrivateCommentData(entity);
+    return privateComment;
+  }
+
+  // only 1 private comment per task
+  public PrivateCommentData getPrivateCommentByTaskID(long taskID) {
+    Filter filter = new FilterPredicate("taskID", FilterOperator.EQUAL, taskID);
+    // private comments should be size = 1
+    ArrayList<PrivateCommentData> privateComments =
+        getPrivateComments(filter, NO_QUERY_LIMIT, NO_QUERY_SORT);
```
Tzirliya, line 45: Maybe run your own query here and use `asSingleEntity()`?

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
```diff
@@ -0,0 +1,102 @@
...
+    return privateComments.get(0);
+  }
+
+  public ArrayList<PrivateCommentData> getPrivateComments(
+      int limit, String sortBy, String sortDirection) {
+    SortPredicate sort =
+        new SortPredicate(sortBy, SortDirection.valueOf(sortDirection.toUpperCase()));
+    return getPrivateComments(NO_QUERY_FILTER, limit, sort);
+  }
```
Tzirliya, line 54: Is there a use case for when users would want to sort the private comment? Wouldn't a user only be possibly sorting their tasks and the private comment would just be displayed underneath/beside each task? Unless I'm misunderstanding how the private comments will be displayed?

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
```diff
@@ -0,0 +1,102 @@
...
+    this.datastore = datastore;
+  }
+
+  public Long addPrivateComment(PrivateCommentData privateComment) {
+    Entity entity = privateComment.toEntity();
+    Key key = datastore.put(entity);
+    long commentID = key.getId();
+    privateComment.setCommentID(commentID);
+    return commentID;
+  }
```
Tzirliya, line 27: Would this be able to add multiple private comments for one userID/taskID combo?

talgsternberg: I was thinking we'd only have 1 private comment per task (since there will only be 1 comment box) but it can constantly be edited and updated.

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
talgsternberg: Yeah you're correct. Just wanted to include this in case the buttons next to the comments looked too clunky and I wanted to display only the comments (and make them buttons to tasks using href)

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
```diff
@@ -0,0 +1,105 @@
...
+  private DatastoreService datastore;
+  private static final Filter NO_QUERY_FILTER = null;
+  private static final int NO_QUERY_LIMIT = Integer.MAX_VALUE;
+  private static final SortPredicate NO_QUERY_SORT = null;
+
+  public PrivateCommentController(DatastoreService datastore) {
+    this.datastore = datastore;
+  }
+
+  public Long addPrivateComment(PrivateCommentData privateComment) {
```
jcespejo, line 21: If we're only allowing for 1 private comment per task, is there anyway we can enforce that?

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentListServlet.java
```diff
@@ -0,0 +1,102 @@
...
+    String[] colors = new String[tasks.size()];
+
+    // build list of private comments and colors
+    ArrayList<PrivateCommentData> privateComments = new ArrayList<>();
+
+    // only fill if the user is viewing their own profile
+    if (userID == currentUser) {
+      for (int i = 0; i < tasks.size(); i++) {
+        Long taskID = tasks.get(i).getTaskID();
+        PrivateCommentData privateComment =
```
jcespejo, line 56: This would do multiple datastore queries to load all private comments for a single user.  Queries can get expensive because of the back and forth between the server and the database.  Anyway to make this more efficient?

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentListServlet.java
```diff
@@ -0,0 +1,102 @@
...
+            privateCommentController.getPrivateCommentByTaskID(taskID);
+        privateComments.add(privateComment);
+        String statusString = tasks.get(i).getStatus().name();
+        if (statusString.equals("COMPLETE")) {
```
jcespejo, line 60: Another option would be to figure out what the colors would be in the JSP since you'll be looping through the comments in the JSP anyways.

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentListServlet.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+
+    // Set attributes of request; retrieve in jsp with
+    // request.setAttribute("userTasks", tasks); I don't need since I can pass from /user-profile
+    request.setAttribute("colors", colors);
+    request.setAttribute("privateComments", privateComments);
+
+    // Load jsp for user page
+    request.getRequestDispatcher("user-private-comments.jsp").forward(request, response);
```
jcespejo, line 74: Are these going to a new page?  I thought these were going to be displayed in the User Profile "task checklist"?

jcespejo (dismissed):
lgtm

src/main/java/com/rtb/projectmanagementtool/user/UserSettingsServlet.java
```diff
@@ -97,7 +97,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     auth.loginUser(request, response);
     Long userLoggedInId = auth.whichUserIsLoggedIn(request, response);
 
-    if (userLoggedInId == /*No user found*/ -1l) {
+    if (userLoggedInId == /*No user found*/ -1) { // changed from -1l
```
jcespejo, line 100: AuthOps.NO_LOGGED_IN_USER

jcespejo (dismissed):
lgtm

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentListServlet.java
talgsternberg: I'm going to do what Tzirliya did with list-tasks.jsp and call that in user_profile.jsp

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
talgsternberg: Not sure. Any recommendations?

jcespejo: @Tzirliya thoughts?

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentListServlet.java
talgsternberg: I'm not sure. I like the loop here because it gets each comment in the order that the tasks will be loaded as well. Would you suggest modifying the .getPrivateCommentByTaskID method? If so, I'm not exactly sure what would change.

jcespejo: One thought would be to load all private comments owned by a single user.  If the assumption is these will only be viewable by the current user in their own user profile pages, this would mean loading all for the user would give you all the records you'll need.

src/main/java/com/rtb/projectmanagementtool/private-comments/PrivateCommentController.java
Tzirliya: Maybe add a method that queries and filters for a privateComment with the current userID / taskID combination. If it returns null, create a new privateCommentData and add it to datastore, otherwise just update the retrieved entity.

PR #82, opened 2020-08-20
Add Edit Features to Task Page

Tzirliya:
Implements some suggestions from previous pull request.

Comments can be edited. JavaScript is used to transform the comment into a form so users can edit the comment without reloading the page. The task description can be edited. It also uses JavaScript to do so.

Messages are shown when hovering over the edit and delete symbol buttons.

I renamed the task.js file to task-old.js file and created a new task.js file that's used when editing the Task Page.

CSS is updated.


src/main/webapp/scripts/task.js
```diff
@@ -1,184 +1,161 @@
...
  */
-function removeUser() {
-  const getParams = new URLSearchParams(location.search);
-  const taskID = getParams.get('taskID');
-  const userID = 1; // Get current userID from cookies?
-  const params = new URLSearchParams('taskID=' + taskID + '&userID=' + userID);
-  console.log('/task-remove-user', params);
-  fetch('/task-remove-user', {method: 'post', body: params})
-      .then(() => getTaskInfo());
+function editComment(commentID, taskID) {
```
jcespejo, line 76: This is a good addition, allowing to edit the comments by dynamically building the form.  From here you can take it a step further by instead of having the form submit and redirect the page, you can have it send a fetch() call to the servlet instead.  If servlet reports a success, you can update the DOM to display the new text (which you should already have since the user had just typed it in).  If the servlet reports a failure, you can pop  up an alert() (or div lightbox) then revert back to the previous comment (which you can hold on to in a hidden div). 

This'll allow updating multiple comments one after the other without having to refresh the page in between.

PR #87, opened 2020-08-21
Create Task Tree

Tzirliya:
A Task Tree data structure is created with Java. A TaskController method is creating to load all the tasks of a project into the tree, given a projectID. A Java Servlet returns the tree. A JSP page is created to traverse the tree and build it in HTML. CSS and JavaScript is added to display the collapsible tree in a pop-up on the Project Page and the Task Page.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
@@ -7,6 +7,7 @@ import com.rtb.projectmanagementtool.user.*;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 
 /** Class controlling the TaskData object. */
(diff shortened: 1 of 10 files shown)


src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -224,6 +225,40 @@ public TaskData getTaskByID(long taskID) {
     return ancestors;
   }
 
+  public ArrayList<TaskTreeData> getTaskTree(long projectID) {
+    ArrayList<TaskTreeData> taskTree = new ArrayList<>();
+    Filter filter = new FilterPredicate("projectID", FilterOperator.EQUAL, projectID);
+    ArrayList<TaskData> tasks = getTasks(filter, NO_QUERY_LIMIT, NO_QUERY_SORT);
+    Iterator<TaskData> iterator = tasks.iterator();
+    while (iterator.hasNext()) {
```
jcespejo, line 233: This part find the root?  I would suggest moving this into buildTree as well, so all tree building related logic (finding the root + adding leaves) is encapsulated in that single method

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -224,6 +225,40 @@ public TaskData getTaskByID(long taskID) {
...
+      return;
+    }
+    ArrayList<TaskTreeData> nextTreeLayer = new ArrayList<>();
+    for (TaskTreeData taskTreeNode : taskTree) {
+      Iterator<TaskData> iterator = tasks.iterator();
+      while (iterator.hasNext()) {
+        TaskData task = iterator.next();
+        if (task.getParentTaskID() == taskTreeNode.getTask().getTaskID()) {
+          nextTreeLayer.add(taskTreeNode.addSubtask(task));
+          iterator.remove();
```
jcespejo, line 255: Repeatedly removing items from the list could get expensive since it'll need to shift all items on the right one step to the left.  Consider building your tree without having to shift the list repeatedly.

src/main/java/com/rtb/projectmanagementtool/task/TaskTreeData.java
```diff
@@ -0,0 +1,64 @@
+package com.rtb.projectmanagementtool.task;
+
+import java.util.ArrayList;
+
+/** Class containing task data. */
+public final class TaskTreeData {
```
jcespejo, line 6: nit:  if we're using "Data" as a suffix for database data models, we should avoid that name here (since the tree itself is not stored in the database).

src/main/java/com/rtb/projectmanagementtool/task/TaskTreeData.java
```diff
@@ -0,0 +1,64 @@
...
+  /** Enum containing status options for a task. */
+  public enum Status {
```
jcespejo, line 8: Same enums as the ones in TaskData?  We should avoid duplicating.

src/main/java/com/rtb/projectmanagementtool/task/TaskTreeData.java
```diff
@@ -0,0 +1,64 @@
...
+    COMPLETE,
+    INCOMPLETE
+  }
+
+  private TaskData task;
+  private ArrayList<TaskTreeData> subtasks;
+
+  public TaskTreeData(TaskData task, ArrayList<TaskTreeData> subtasks) {
+    this.task = task;
```
jcespejo, line 17: In this case since you have both the parent and the child tasks in the same structure, it would be clearer if you could explicitly label the parentTask

src/main/webapp/scripts/project.js
```diff
@@ -16,4 +16,49 @@ function showAddUserForm() {
...
+    if (event.target == popup) {
+      popup.style.display = 'none';
+    }
+  };
+}
+
+/**
+ * Body onload function for Project Page.
+ */
+function loadProjectPage() {
```
jcespejo, line 61: nit:  initEventListeners?  since we're not loading anything more from this point.

PR #88, opened 2020-08-24
Create Task Blocker Feature

Tzirliya:
A simple version of the Task Blocker feature is created. TaskBlocker data and controller classes are created. A JSP page is added to choose a task and its blocker from two Task Trees and post a TaskBlocker entity via the TaskBlocker servlet to datastore.
Suggestions regarding the Task Tree from the previous pull request are implemented.
There's an icon at the top of each Task Page that shows if a task is blocked and if so, it links to the blocker tasks when hovering over the icon.
When setting a task's status as complete, search to see if it has any blockers first.


src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -2,17 +2,99 @@
...
+  // Add methods
+
+  public void addTaskBlocker(long taskID, long blockerID) {
+    addTaskBlocker(new TaskBlockerData(taskID, blockerID));
+  }
+
+  public void addTaskBlocker(TaskBlockerData taskBlocker) {
+    // Get task
+    TaskData task;
+    try {
```
jcespejo, line 30: These two can be combined to build a single query that returns two tasks

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -2,17 +2,99 @@
...
+      System.out.println("Cannot find task with provided blockerID");
+      return;
+    }
+    // Add taskBlocker
+    if (task.getStatus() != Status.COMPLETE) {
+      TransactionOptions options = TransactionOptions.Builder.withXG(true);
+      Transaction transaction = datastore.beginTransaction(options);
+      try {
+        datastore.put(taskBlocker.toEntity());
+        assert containsCycle() == false;
```
jcespejo, line 50: Look into checking for cycles first, before adding the record in the datastore.  The transaction->add->rollback process could be expensive and unnecessary if we could very beforehand.

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -2,17 +2,99 @@
...
+    return false;
+  }
+
+  // Get methods
+
+  public HashSet<TaskBlockerData> getAllTaskBlockers() {
+    return getTaskBlockers(NO_FILTER);
+  }
+
+  public ArrayList<TaskData> getTaskBlockerTasks(long taskID) {
```
jcespejo, line 73: nit: getBlockersForTask?

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -2,17 +2,99 @@
...
+    HashSet<TaskBlockerData> taskBlockers = getTaskBlockers(taskID);
+    ArrayList<TaskData> tasks = new ArrayList<>();
+    for (TaskBlockerData taskBlocker : taskBlockers) {
+      tasks.add(taskController.getTaskByID(taskBlocker.getBlockerID()));
```
jcespejo, line 77: Queries in loops could get expensive.  Maybe look into building a list of keys and doing 1 big query.

PR #89, opened 2020-08-24
textarea with private comments

talgsternberg:
private comment that can be updated/displayed in textareas next to task buttons in red or green based on completion status


src/main/webapp/user-private-comments.jsp
```diff
@@ -23,10 +24,10 @@
         <%}%>
           <%=task.getName()%>
         </button>
-        <form class="inline" id="add-private-comment-post-form" action="/user-profile" method="POST">
-          <input type="hidden" id="add-private-comment-task-input" name="taskID" value="<%=task.getTaskID()%>">
+        <form action="/user-profile" method="POST">
```
jcespejo, line 27: You have multiples of these on the page I believe (one per task), but all of them have the inputs/textarea have the same name "taskID" and "message".  I'm not sure how HTTP would handle this, whether it'll send a list of if it'll send just the first one.  Try to look at the Chrome Dev Console so see what the request looks like.


Also since you're submitting all of them into the same user-profile servlet, that means it'll try to resave all messages into datastore, which would then be inefficient if there are comments that were not updated.

Try to see if you can build it so each comment would submit (and save in datastore) individually.

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -109,24 +108,29 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
-      // Create PrivateCommentData object
-      PrivateCommentData privateComment = new PrivateCommentData(taskID, userID, message);
+      // Add updates to datastore
+      pcController.updatePrivateComment(pcData);
+    }
 
-      // Add private comment to datastore
-      PrivateCommentController privateCommentController = new PrivateCommentController(datastore);
-      privateCommentController.addPrivateComment(privateComment);
+    // otherwise, create a new pcData and put that in ds
```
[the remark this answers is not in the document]
talgsternberg: Update: it seems this is being hit every time

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -109,14 +109,16 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
       throws ServletException, IOException {
 
     // get parameters
-    long taskID = Long.parseLong(request.getParameter("taskID"));
```
talgsternberg: this is null even though I thought I was passing it from user-private-comments.jsp. the task (which I printed out in the jsp) is not null, so this is just not being passed.

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -109,7 +109,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
       throws ServletException, IOException {
 
     // get parameters
-    TaskData task = (TaskData) request.getParameter("task")
+    TaskData task = (TaskData) request.getAttribute("task");
```
talgsternberg, line 112: this is null even though I thought I was passing it from user-private-comments.jsp. the task (which I printed out in the jsp) is not null, so this is just not being passed.

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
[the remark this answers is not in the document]
godsfavours:
```
long taskID = Long.parseLong(request.getParameter("taskID-" + task.getName()));
```
taskID is probably null because that specific parameter doesn't exist. In jsp you send in 
```
name="taskID-"<%=task.getName()%>
```
Only the 'taskID-' part is sent to the servlet. Putting <%=task.getName()%> inside the quotation mark should do:
```
name="taskID-<%=task.getName()%>"
```

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
godsfavours:
```
<%request.setAttribute("task", task);%>
```
I'm not entirely sure if this is the cause, but this line in the jsp is not inside of the form. Might be why it's null. Instead of trying to parse TaskData from the "task" parameter, and since you've already sent in the taskID, you could use taskController to get the TaskData

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -109,24 +108,29 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
-    TaskController taskController = new TaskController(datastore);
-    ArrayList<TaskData> tasks = taskController.getTasksByUserID(userID);
+    // new controller for methods
+    PrivateCommentController pcController = new PrivateCommentController(datastore);
 
-    // for each task, get params and create private comments
-    for (TaskData task : tasks) {
-      long taskID = Long.parseLong(request.getParameter("taskID"));
-      String message = request.getParameter("message");
+    // if a pcomment already exists with this taskID. Get it and update it.
```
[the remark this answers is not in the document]
jcespejo: It looks like it should.  If not, check if the commentID (entity key) is correctly filled in when you're writing to the datastore.  It should be somewhere around here:
https://github.com/googleinterns/step171-2020/blob/master/src/main/java/com/rtb/projectmanagementtool/privatecomment/PrivateCommentData.java#L43

If that's not filled in, check if you're filling it in when pulling the comments from the datastore

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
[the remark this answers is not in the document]
jcespejo: If this is getting hit every time, it means `pcController.getPrivateCommentByTaskID(taskID)` either is not finding the record (check the query and filters), or it's not filling in the key and/or taskID correctly (so looking up the taskID is not returning results)

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
jcespejo: You don't have an input with name="task" in your form, so you wouldn't get an attribute with name "task" in your request.  Even if you add the taskData as an input in your form, HTTP requests will only send you string data, which means you'll need to convert to taskData somehow.

It also might not be a good idea to pass the whole task in the request if you don't need the whole task (eg. you probably don't need the task description).  See if you can instead just pass the taskID into the request.

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -98,24 +108,33 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
   public void doPost(HttpServletRequest request, HttpServletResponse response)
       throws ServletException, IOException {
 
-    Long userID = auth.whichUserIsLoggedIn(request, response);
+    // get parameters
+    TaskData task = (TaskData) request.getAttribute("task");
+    long taskID = Long.parseLong(request.getParameter("taskID-" + task.getName()));
```
jcespejo, line 113: Try not to key the request params by the task Name, since the name is not guaranteed to be unique.  Try to use the taskID instead.

src/main/java/com/rtb/projectmanagementtool/user/UserProfileServlet.java
```diff
@@ -57,37 +57,47 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
 
-    // build map of task to pcomment
+    // build map of taskID to pcomment
     Map<Long, PrivateCommentData> privateCommentsMap = new HashMap<Long, PrivateCommentData>();
 
-    // preset each comment for each task empty
-    for (TaskData task : tasks) {
-      privateCommentsMap.put(
-          task.getTaskID(), new PrivateCommentData(task.getTaskID(), userID, ""));
+    // preset each comment if every task is empty
```
jcespejo, line 82: I think this is the right idea, creating a blank comment object if the task does not have a comment yet so you won't have to deal with nulls later on.  But, right now you're only doing the pre-fill if all comments for all tasks are missing.  For example if they have 4 tasks with no comments, initially this will work great to pre-fill with black comments.  But, once they add a comment for one of those 4 tasks, then the next time they go to this page only the comments for that 1 task will be filled (with legit data this time), and then you're back dealing with nulls on your jsp

talgsternberg: I think my most recent commit will fix this issue

PR #90, opened 2020-08-24
Add Algorithm to Check for Cycles

Tzirliya:
A path-finding algorithm is added to check if there will be a cycle if a TaskBlocker entity is added to datastore. Unit tests are added.
Suggestions from the previous pull request are implemented.

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
+++ b/src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
@@ -6,6 +6,9 @@ import com.rtb.projectmanagementtool.task.*;
 import com.rtb.projectmanagementtool.task.TaskData.Status;
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /** Class controlling the TaskBlockerData object. */
 public final class TaskBlockerController {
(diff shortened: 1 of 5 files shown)


src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
+    taskBlocker.setTaskBlockerID(datastore.put(taskBlocker.toEntity()).getId());
   }
 
-  private boolean containsCycle() {
-    // TODO: Implement cycle detection algorithm
+  private boolean containsPath(long start, long end) {
+    HashSet<TaskBlockerData> taskBlockers = getAllTaskBlockers();
+    Map<Long, Boolean> taskBlockersVisited =
+        taskBlockers
+            .stream()
```
Tzirliya, line 59: The Java presubmit test is failing because it wants to combine these two lines onto one line, but they automatically split up again whenever the code is run.

jcespejo: The coveo plugin in pom.xml is using version 2.5.0 which was released in 2018, while the java formatter used by the [presubmit checks](https://github.com/googleinterns/step171-2020/blob/master/.github/workflows/javalint.yaml) uses version 1.8, which was released in May 2020.  Maybe a rule was updated recently and the presubmit is running newer rules, whole the pom.xml plugin is running older rules.  Try a newer version of the coveo plugin:  https://github.com/coveooss/fmt-maven-plugin/releases

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
+      return;
     }
+    // Add the TaskBlocker
+    taskBlocker.setTaskBlockerID(datastore.put(taskBlocker.toEntity()).getId());
   }
 
-  private boolean containsCycle() {
-    // TODO: Implement cycle detection algorithm
+  private boolean containsPath(long start, long end) {
+    HashSet<TaskBlockerData> taskBlockers = getAllTaskBlockers();
```
jcespejo, line 56: This would load all blockers in the datastore, although I believe we said before they're only allowed to block tasks within the same project, correct?

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
+            .stream()
+            .collect(Collectors.toMap(x -> x.getTaskID(), x -> false, (x1, x2) -> x1));
+    LinkedList<Long> queue = new LinkedList<>();
+    taskBlockersVisited.put(start, true);
+    queue.add(start);
+    Long blockerID;
+    ArrayList<TaskBlockerData> toRemove = new ArrayList<>();
+    while (queue.size() != 0) {
+      start = queue.poll();
+      for (TaskBlockerData taskBlocker : taskBlockers) {
```
jcespejo, line 68: If you have blocker chains 1->2->3->4->5->6 and 8->9, when looking for a path from 6->2, this loop will continually visit the 8->9 blocker even though this graph is disconnected from the nodes we care about.  Thoughts on optimizing?

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
+          }
+          if (taskBlockersVisited.containsKey(blockerID)
+              && taskBlockersVisited.get(blockerID) == false) {
+            taskBlockersVisited.put(blockerID, true);
+            queue.add(blockerID);
+          }
+          toRemove.add(taskBlocker);
+        }
+      }
+      taskBlockers.removeAll(toRemove);
```
jcespejo, line 82: deleting from containers could get expensive.  Depending on the language, this would either resize the container and/or rehash the keys.  It would be best to avoid deleting if possible.

iiuc, you're deleting the nodes that  you've already visited.  In which case, checking the visited map could be another option.

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
+    ArrayList<TaskBlockerData> toRemove = new ArrayList<>();
+    while (queue.size() != 0) {
+      start = queue.poll();
+      for (TaskBlockerData taskBlocker : taskBlockers) {
+        if (taskBlocker.getTaskID() == start) {
+          blockerID = taskBlocker.getBlockerID();
+          if (blockerID == end) {
+            return true;
+          }
+          if (taskBlockersVisited.containsKey(blockerID)
```
jcespejo, line 74: containsKey, get, and put all does separate lookups through the map.  Any way to optimize by combining lookups?

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
+      System.out.println("Cannot block a task if it would create a cycle.");
+      return;
     }
+    // Add the TaskBlocker
+    taskBlocker.setTaskBlockerID(datastore.put(taskBlocker.toEntity()).getId());
   }
 
-  private boolean containsCycle() {
-    // TODO: Implement cycle detection algorithm
+  private boolean containsPath(long start, long end) {
```
jcespejo, line 55: This would query the datastore + "build" and traverse the graph **every time** you need to add a new blocker.  (build in quotes because the algorithm you're using essentially builds and traverses the graph, just doesn't store the graph into a data structure).  Thoughts on optimizing?

Also, we don't have the feature at the moment, but if we build a "show all blockers of blockers" feature, optimization here could help with that as well.

src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -25,58 +29,88 @@ public void addTaskBlocker(long taskID, long blockerID) {
...
-        if (transaction.isActive()) {
-          transaction.rollback();
-        }
-      }
-    } else {
-      System.out.println("Cannot block a task set to COMPLETE.");
+    // Ensure a cycle wouldn't be created if the TaskBlocker is added
+    if (containsPath(taskBlocker.getBlockerID(), taskBlocker.getTaskID())) {
+      System.out.println("Cannot block a task if it would create a cycle.");
+      return;
```
jcespejo, line 49: This should return a status and let the user know if adding the blocker failed.

PR #91, opened 2020-08-25
Final updates before working on chat feature

godsfavours:
Implemented some basic features relating to the project page in order to have a better functioning product. Apologies if PR is big; I was having trouble with js presubmit warnings and it was easier to submit this all in one go.

Most notable changes/additions:
- home page & project page styled
- added modals to display popups
- added 'email' field to UserData to allow for adding users to projects via email (will need to delete all User entitties in _/ah/admin to avoid errors)
- added 'isCompleted' field to ProjectData to allow to implement completing projects feature (will need to delete all Project entities in _/ah/admin)
- added 'edit project details' feature to change the name/description of a project
- added project removal functionality


src/main/java/com/rtb/projectmanagementtool/project/CompleteProjectServlet.java
```diff
@@ -0,0 +1,73 @@
...
+      if (task.getStatus() == Status.INCOMPLETE) {
+        response
+            .getWriter()
+            .println(
+                generateResponse(/*message*/ "Task(s) incomplete. Unable to complete project."));
+        return;
+      }
+    }
+
+    // All project tasks are incomplete
```
Tzirliya, line 57: nit: shouldn't it be `complete`?

src/main/java/com/rtb/projectmanagementtool/task/TaskController.java
```diff
@@ -286,6 +286,13 @@ private void buildTree(ArrayList<TaskTreeNode> taskTree, ArrayList<TaskData> tas
   }
 
   // Delete methods
+  public void deleteTasksGivenTaskData(ArrayList<TaskData> tasks) {
+    ArrayList<Long> taskIDs = new ArrayList<Long>();
+    for (TaskData task : tasks) {
+      taskIDs.add(task.getTaskID());
+    }
```
Tzirliya, line 293: You can use `getTaskIDsFromTasks()` which is already in the TaskController.

src/test/java/com/rtb/projectmanagementtool/user/UserControllerTest.java
```diff
@@ -45,11 +45,11 @@
...
   // UserData
   private static final UserData user1 =
-      new UserData(userID1, AuthID1, name1, year1, majors1, skills1, totalCompTasks1);
+      new UserData(userID1, AuthID1, /*email*/ "", name1, year1, majors1, skills1, totalCompTasks1);
   private static final UserData user2 =
-      new UserData(userID2, AuthID2, name2, year2, majors2, skills2, totalCompTasks2);
+      new UserData(userID2, AuthID2, /*email*/ "", name2, year2, majors2, skills2, totalCompTasks2);
   private static final UserData user3 =
-      new UserData(userID3, AuthID3, name3, year3, majors3, skills3, totalCompTasks3);
+      new UserData(userID3, AuthID3, /*email*/ "", name3, year3, majors3, skills3, totalCompTasks3);
```
Tzirliya, line 52: nit: Why not add email as a variable above, like the rest of the attributes?

godsfavours:
> Please don't store the UserService (auth) email into the datastore. You can instead add an email field to UserData and fill it in with fake emails.
> 
> Storing actual user-identifiable info usually would require Legal and Privacy approval, which we don't want to deal with right now.
> 
> Also, the idea behind the Auth and UserData separation was so we can easily decouple them, and that the only connection between them is the AuthID. Hypothetically we can switch to a different auth system (even a homebrew one) that provides some sort of ID, and everything else would work as is. Adding a dependency on email would break this decoupling.

done

src/main/java/com/rtb/projectmanagementtool/auth/LoginServlet.java
```diff
@@ -19,21 +17,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
-    // Don't view login page if user is logged in
-    if (auth.whichUserIsLoggedIn(request, response) != AuthOps.NO_LOGGED_IN_USER) {
-      response.sendRedirect("/home");
-      return;
+    String authID = auth.getAuthID();
+    // If user has no auth id,
+    if (authID != null) {
+      auth.loginUser(request, response); // try to log in
+      UserController userController = new UserController(datastore);
+      UserData user = userController.getUserByAuthID(authID);
```
jcespejo, line 25: same calls

src/main/java/com/rtb/projectmanagementtool/project/DeleteProjectServlet.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/delete-project")
+public class DeleteProjectServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
```
jcespejo, line 16: GET shouldn't forward to POST.  They're sematically different and should handle different types of requests.

src/main/java/com/rtb/projectmanagementtool/user/UserData.java
```diff
@@ -24,6 +25,7 @@
 
   private long userID;
   private String AuthID; // this will be the ID from API
+  private String inviteCode; // code used to add user to a project
```
jcespejo, line 28: Will the project admin (professor) be required to type these in to find students?  It would be unlikely for a professor to know these, unless if they have some other reference table to look at that would give the codes for students names.  In which case maybe we can provide that list (of users) for them?

Email would have been a better approach than inviteCode (with made up emails for now).

PR #92, opened 2020-08-26
Fix UserID Issue and Add More Stylistic Stuff

talgsternberg:


src/main/webapp/user-private-comments.jsp
```diff
@@ -16,20 +18,20 @@
...
+        <%request.setAttribute("clickable", false);%>
+        <jsp:include page="task-status-checkmark.jsp"/>
+        <p class="inline"><%=task.getDescription()%></p>
         <form action="/user-profile" method="POST">
           <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
           <br>
           <textarea style="width:300px; height:100px;" type="text" name="message-<%=task.getTaskID()%>" rows="6" cols="50"><%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%></textarea>
           <br>
-          <button class="deep-button" type="submit" class="deep-button">Update</button>
+          <button class="deep-button" style="font-size: 12px;" type="submit" class="deep-button">Update Comment</button>
```
jcespejo, line 34: Please move into the css for consistency.  If it helps, you can have 2 classes in the same tag 
eg `class="deep-button submit-button"` 

then reference them in the css file separately like:
```
.deep-button{...}
.submit-button{
  font-size:  12px;
}
```

PR #93, opened 2020-08-27
Create TaskBlocker Graph that's stored in Memcache

Tzirliya:
Implemented optimization suggestions from previous pull request.
Raise an exception with an error message when a user adds an invalid task blocker combo, which the user sees in an alert.
A graph class is created to hold task blocker info. It's stored in memcache.


src/main/java/com/rtb/projectmanagementtool/taskblocker/TaskBlockerController.java
```diff
@@ -2,89 +2,138 @@
...
     if (containsPath(taskBlocker.getBlockerID(), taskBlocker.getTaskID())) {
-      System.out.println("Cannot block a task if it would create a cycle.");
-      return;
+      throw new TaskBlockerException("Cannot block a task if it would create a cycle");
     }
     // Add the TaskBlocker
     taskBlocker.setTaskBlockerID(datastore.put(taskBlocker.toEntity()).getId());
   }
 
   private boolean containsPath(long start, long end) {
```
jcespejo, line 65: This can be a method of the graph class, since finding a path could be a valid operation for the graph itself.  One advantage of moving this in the graph: if you switch graph implementations for example from HashMaps to actual Node classes, you wouldn't have to change any of the files that requires checking for paths.

PR #94, opened 2020-08-27
Attempting EditorJS

talgsternberg:
still buggy

### DIFF
--- a/src/main/webapp/user-private-comments.jsp
+++ b/src/main/webapp/user-private-comments.jsp
@@ -5,7 +5,6 @@
 <%@ page import="java.util.*"%>
 
 
-
 <%--Get variables--%>
 <%
     List<TaskData> tasks = (List<TaskData>)(List<?>) request.getAttribute("tasks");
@@ -14,27 +13,53 @@
 %>
 
 <%--HTML--%>
-<ul>
-  <%if (currUser) {%>
-    <h1 id="pc-header">Your Private Task Comments:</h1>
-    <%for (TaskData task : tasks) {%>
-      <li class="task-with-pc">
-        <button class="inline deep-button" type="button" onclick="location.href='task?taskID=<%=task.getTaskID()%>'">
-          <%=task.getName()%>
-        </button>
-        <%request.setAttribute("task", task);%>
-        <%request.setAttribute("clickable", false);%>
-        <jsp:include page="task-status-checkmark.jsp"/>
-        <p class="inline"><%=task.getDescription()%></p>
-        <form action="/user-profile" method="POST">
-          <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
-          <br>
-          <textarea style="width:300px; height:100px;" type="text" name="message-<%=task.getTaskID()%>" rows="6" cols="50"><%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%></textarea>
-          <br>
-          <button class="deep-button" style="font-size: 12px;" type="submit" class="deep-button">Update Comment</button>
-        </form>
-        <br><br>
-      </li>
+<head>
+  <script src="https://cdn.jsdelivr.net/npm/@editorjs/editorjs@latest"></script>
+  <script src="https://cdn.jsdelivr.net/npm/@editorjs/header@latest"></script>
+</head>   
+  <ul>
+    <%if (currUser) {%>
+      <h1 id="pc-header">Your Private Task Comments:</h1>
+      <%for (TaskData task : tasks) {%>
+        
+        <script>
+          function init() {
+            const editor = new EditorJS({
+            holderId : 'box',
+            autofocus: true,
+            data: '<%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>',
+            tools: {
+              header: Header,
+             },
+            });
+          }
+          function save() {
+            editor.save().then((outputData) => {
+              console.log('Article data: ', outputData)
+            }).catch((error) => {
+              console.log('Saving failed: ', error)
+            });  
+          }
+        </script>
+        
+        <li class="task-with-pc">
+          <button class="inline deep-button" type="button" onclick="location.href='task?taskID=<%=task.getTaskID()%>'">
+            <%=task.getName()%>
+          </button>
+          <%request.setAttribute("task", task);%>
+          <%request.setAttribute("clickable", false);%>
+          <jsp:include page="task-status-checkmark.jsp"/>
+          <p class="inline"><%=task.getDescription()%></p>
+          <form onload="init()" action="/user-profile" method="POST">
+            <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
+            <br>
+            <div id='box' style="width:300px; height:100px;" type="text" name="message-<%=task.getTaskID()%>"></div>
+            <br>
+            <button onlclick="save()" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
+          </form>
+          <br><br>
+        </li>
+      <%}%>
     <%}%>
-  <%}%>
-</ul>
+  </ul>
+</body>
--- a/src/main/webapp/user-private-comments.jsp
+++ b/src/main/webapp/user-private-comments.jsp
@@ -42,7 +42,6 @@
       <%if (currUser) {%>
         <h1 id="pc-header">Your Private Task Comments:</h1>
         <%for (TaskData task : tasks) {%>
-          <%System.out.println(task);
           <li class="task-with-pc">
             <button class="inline deep-button" type="button" onclick="location.href='task?taskID=<%=task.getTaskID()%>'">
               <%=task.getName()%>
@@ -51,10 +50,10 @@
             <%request.setAttribute("clickable", false);%>
             <jsp:include page="task-status-checkmark.jsp"/>
             <p class="inline"><%=task.getDescription()%></p>
-            <form onload="init(box-<%=task.getTaskID()%>, <%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim());" action="/user-profile" method="POST">
+            <form action="/user-profile" method="POST">
               <input type="hidden" name="taskID" value="<%=task.getTaskID()%>"> 
               <br>
-              <div id="box-<%=task.getTaskID()%>" type="text"></div>
+              <div id="box-<%=task.getTaskID()%>" onload="init('box-<%=task.getTaskID()%>', '<%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>')" type="text"></div>
               <br>
               <input type="hidden" name="message-<%=task.getTaskID()%>" value="document.getElementById('box-<%=task.getTaskID()%>').value(outputData);"</input>
               <button onclick="save()" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
--- a/src/main/webapp/user-private-comments.jsp
+++ b/src/main/webapp/user-private-comments.jsp
@@ -20,7 +20,8 @@
   </head>
   <script>
     function init(privateCommentID, jsData) {
-      const editor = new EditorJS({
+        console.log("here!");
+        const editor = new EditorJS({
         holderId : privateCommentID,
         autofocus: true,
         data: jsData,
@@ -28,6 +29,7 @@
           header: Header,
         },
       });
+      console.log("here 2 !");
     }
     function save() {
       editor.save().then((outputData) => {
--- a/src/main/webapp/user-private-comments.jsp
+++ b/src/main/webapp/user-private-comments.jsp
@@ -53,8 +53,11 @@
             <form action="/user-profile" method="POST">
               <input type="hidden" name="taskID" value="<%=task.getTaskID()%>"> 
               <br>
-              <div id="box-<%=task.getTaskID()%>" onload="init('box-<%=task.getTaskID()%>', '<%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>')" type="text"></div>
-              <br>
+              <%
+                PrivateCommentData pc = (PrivateCommentData) (privateCommentsMap.get(task.getTaskID()));
+                String message = (String) pc.getMessage().trim();
+              %>
+              <div id="box-<%=task.getTaskID()%>" onload="init('box-<%=task.getTaskID()%>', '<%=message%>')" type="text"></div>
               <input type="hidden" name="message-<%=task.getTaskID()%>" value="document.getElementById('box-<%=task.getTaskID()%>').value(outputData);"</input>
               <button onclick="save()" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
             </form>
--- a/src/main/webapp/user-private-comments.jsp
+++ b/src/main/webapp/user-private-comments.jsp
@@ -21,13 +21,13 @@
     <%if (currUser) {%>
       <h1 id="pc-header">Your Private Task Comments:</h1>
       <%for (TaskData task : tasks) {%>
-        
+        var jsData = <%=var jsData = (privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>;
         <script>
           function init() {
             const editor = new EditorJS({
             holderId : 'box',
             autofocus: true,
-            data: '<%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>',
+            data: jsData,
             tools: {
               header: Header,
              },
@@ -53,7 +53,7 @@
           <form onload="init()" action="/user-profile" method="POST">
             <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
             <br>
-            <div id='box' style="width:300px; height:100px;" type="text" name="message-<%=task.getTaskID()%>"></div>
+            <div id='box' type="text" name="message-<%=task.getTaskID()%>"></div>
             <br>
             <button onlclick="save()" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
           </form>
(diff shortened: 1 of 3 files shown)


src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,53 @@
...
+    <%if (currUser) {%>
+      <h1 id="pc-header">Your Private Task Comments:</h1>
+      <%for (TaskData task : tasks) {%>
+        
+        <script>
+          function init() {
+            const editor = new EditorJS({
+            holderId : 'box',
+            autofocus: true,
+            data: '<%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>',
```
talgsternberg: this is not working to load from java. Is there a different way to do it? I looked at a couple of examples and this seemed like the right way to pass from java-->js

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,53 @@
...
+            tools: {
+              header: Header,
+             },
+            });
+          }
+          function save() {
+            editor.save().then((outputData) => {
```
talgsternberg: would "outputData" be what I want to pass to an <input> to save in Datastore?

talgsternberg: If so, can I just somehow reformat it to java?

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,53 @@
...
-        <form action="/user-profile" method="POST">
-          <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
-          <br>
-          <textarea style="width:300px; height:100px;" type="text" name="message-<%=task.getTaskID()%>" rows="6" cols="50"><%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%></textarea>
-          <br>
-          <button class="deep-button" style="font-size: 12px;" type="submit" class="deep-button">Update Comment</button>
-        </form>
-        <br><br>
-      </li>
+<head>
```
jcespejo: These should go into the HTML's head tag.  I believe user-prive-comment.jsp is inserted into user-profile?  The head tag is probably on user-profile.jsp

src/main/webapp/user-private-comments.jsp
jcespejo: your editorjs setup looks correct I think.  Couple things though:
- `holderId : 'box'`  <- "box" here is the ID of the div you want to convert into an editorjs box.  I'm not seeing where you have a `<div id="box">`
- since you have the setup wrapped in an `init()` function, you'll need to call the function for it to build the editor box
- I see you have the scripts with the init() and save() functions inside the tasks loop.  This will mean it'll create the init() and save() methods multiple times.  JS I think lets you do this (other languages would complain), but it would only keep the last version, meaning the one with the last tasks's info.  What you can do instead is build init() and save() once (probably in the js file), but be flexible enough to be called on for multiple boxes (since you have multiple tasks)

src/main/webapp/user-private-comments.jsp
jcespejo: yep outputData would contain the box data that you'd save into Datastore.  from here, you can then push that into an input right before the form is submitted:
```
document.getElementById('private-comment-12345').value(outputData);
```

src/main/webapp/user-private-comments.jsp
```diff
@@ -21,13 +21,13 @@
     <%if (currUser) {%>
       <h1 id="pc-header">Your Private Task Comments:</h1>
       <%for (TaskData task : tasks) {%>
-        
+        var jsData = <%=var jsData = (privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>;
```
talgsternberg, line 24: update: looked up some stuff and found that I could declare as a var with this syntax. yet, this line is still erroring.

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,53 @@
...
-        <br><br>
-      </li>
+<head>
+  <script src="https://cdn.jsdelivr.net/npm/@editorjs/editorjs@latest"></script>
+  <script src="https://cdn.jsdelivr.net/npm/@editorjs/header@latest"></script>
+</head>   
+  <ul>
+    <%if (currUser) {%>
+      <h1 id="pc-header">Your Private Task Comments:</h1>
+      <%for (TaskData task : tasks) {%>
```
jcespejo: Since you have your "init()" and "save()" methods in a loop, this will create multiple copies of it.  This is ok in Javascript (other languages would complain), but it'll only keep the last version (the one for the last task).

You also have your <div id="box"> inside the loop, which will be a problem for the editorjs since it can't detect which box is the correct box.

What you can do instead:
-  Move init() and save() into the js file, but have the methods be flexible enough to work for the different tasks (maybe pass in a privateCommentId so you know which box to build).
- This will also mean you'll have multiple "editor" objects (one per comment), so you'll need to store that in an array

jcespejo (dismissed):
editorjs setup looks good.  I think the problem is because you have the setup in the tasks loop, and you now have multiples of those.

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,53 @@
...
+        var jsData = <%=var jsData = (privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>;
```
jcespejo: the version you had before was good (the one were you were setting data: directly.  This is good as well, but you'll need to move it into the <script>

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,56 @@
...
+            <%request.setAttribute("task", task);%>
+            <%request.setAttribute("clickable", false);%>
+            <jsp:include page="task-status-checkmark.jsp"/>
+            <p class="inline"><%=task.getDescription()%></p>
+            <form onload="init()" action="/user-profile" method="POST">
+              <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
+              <br>
+              <div id="box" type="text"></div>
+              <br>
+              <input type="hidden" name="message-<%=task.getTaskID()%>" value="<%=document.getElementById('box').value(outputData);%>"</input>
```
talgsternberg: error here: "Invalid character constant". I'm not sure but I think that has to do with "outputData". Perhaps I call this only after submit?

Also I think getting the element ID shouldn't be an issue (because the box id changes for every task in the loop) or should I name the ID box-<%task.getTaskID()%> ?

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,56 @@
...
+  <head>
+    <script src="https://cdn.jsdelivr.net/npm/@editorjs/editorjs@latest"></script>
+    <script src="https://cdn.jsdelivr.net/npm/@editorjs/header@latest"></script>
+  </head>
+  <script>
+    function init() {
+      const editor = new EditorJS({
+        holderId : 'box',
+        autofocus: true,
+        data: jsData,
```
talgsternberg: Is it okay to leave this as an undefined var here but define it every iteration of my loop below?

src/main/webapp/user-private-comments.jsp
```diff
@@ -13,53 +13,56 @@
...
-          </form>
-          <br><br>
-        </li>
+<html>
+  <head>
+    <script src="https://cdn.jsdelivr.net/npm/@editorjs/editorjs@latest"></script>
+    <script src="https://cdn.jsdelivr.net/npm/@editorjs/header@latest"></script>
+  </head>
+  <script>
+    function init() {
```
jcespejo: Not here, but would need to be [here](https://github.com/googleinterns/step171-2020/blob/master/src/main/webapp/user_profile.jsp#L22) since this whole jsp is included into the user profile page.  If you navigate to user-profile and View Source, you'll see the extra html/head/body tags you have here will create a malformed html

src/main/webapp/user-private-comments.jsp
jcespejo: you'd need to pass the data into the function:
```
function init(priveCommentId, jsData) {...}
```

then in your loop below:
```
<form onload="init(priveCommentId, data)"
```

src/main/webapp/user-private-comments.jsp
jcespejo: > Also I think getting the element ID shouldn't be an issue (because the box id changes for every task in the loop) or should I name the ID box-<%task.getTaskID()%> ?

the editorjs will need divs with different ids, since "id=box" will match a bunch at once.


> <input type="hidden" name="message-<%=task.getTaskID()%>" value="<%=document.getElementById('box').value(outputData);%>"</input>

I'm not sure about this one.  It's mixing jsp/java (the `<%` tags) with javascript (document.getElementById).  Although you're building them both here at the same time, they are executed separately (jsp/java on the server, javascript on the client)

src/main/webapp/user-private-comments.jsp
```diff
@@ -51,13 +51,13 @@
             <%request.setAttribute("clickable", false);%>
             <jsp:include page="task-status-checkmark.jsp"/>
             <p class="inline"><%=task.getDescription()%></p>
-            <form onload="init()" action="/user-profile" method="POST">
-              <input type="hidden" name="taskID" value="<%=task.getTaskID()%>">
+            <form onload="init(box-<%=task.getTaskID()%>, <%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim());" action="/user-profile" method="POST">
```
talgsternberg: This line and the <input> line below it cause all the issues on the page with loading task. I'm going to do some more specific prints to figure out the exact issues.

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,56 @@
...
+        console.log('Saving failed: ', error)
+      });  
+    }
+  </script> 
+  <body>  
+    <ul>
+      <%if (currUser) {%>
+        <h1 id="pc-header">Your Private Task Comments:</h1>
+        <%for (TaskData task : tasks) {%>
+          <%System.out.println(task);
```
jcespejo: Missing a closing tag?

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,56 @@
...
+          <li class="task-with-pc">
+            <button class="inline deep-button" type="button" onclick="location.href='task?taskID=<%=task.getTaskID()%>'">
+              <%=task.getName()%>
+            </button>
+            <%request.setAttribute("task", task);%>
+            <%request.setAttribute("clickable", false);%>
+            <jsp:include page="task-status-checkmark.jsp"/>
+            <p class="inline"><%=task.getDescription()%></p>
+            <form onload="init(box-<%=task.getTaskID()%>, <%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim());" action="/user-profile" method="POST">
```
jcespejo: also missing a closing tag I believe

jcespejo: What out with this, after you get past the jsp error your js will probably error out since box-12345 is not quoted.

src/main/webapp/user-private-comments.jsp
```diff
@@ -51,10 +50,10 @@
             <%request.setAttribute("clickable", false);%>
             <jsp:include page="task-status-checkmark.jsp"/>
             <p class="inline"><%=task.getDescription()%></p>
-            <form onload="init(box-<%=task.getTaskID()%>, <%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim());" action="/user-profile" method="POST">
+            <form action="/user-profile" method="POST">
               <input type="hidden" name="taskID" value="<%=task.getTaskID()%>"> 
               <br>
-              <div id="box-<%=task.getTaskID()%>" type="text"></div>
+              <div id="box-<%=task.getTaskID()%>" onload="init('box-<%=task.getTaskID()%>', '<%=(privateCommentsMap.get(task.getTaskID())).getMessage().trim()%>')" type="text"></div>
```
talgsternberg: this line has a null pointer somewhere. I'll play around with prints and readjusting quotation marks, etc...

src/main/webapp/user-private-comments.jsp
```diff
@@ -14,27 +13,58 @@
...
-  <%}%>
-</ul>
\ No newline at end of file
+<html>
+  <head>
+    <script src="https://cdn.jsdelivr.net/npm/@editorjs/editorjs@latest"></script>
+    <script src="https://cdn.jsdelivr.net/npm/@editorjs/header@latest"></script>
+  </head>
+  <script>
+    function init(privateCommentID, jsData) {
```
jcespejo: try to add prints in this method so you can follow if it gets called or not.  `console.log("...")` I believe?

src/main/webapp/user-private-comments.jsp
```diff
@@ -56,11 +57,10 @@
               <input type="hidden" name="taskID" value="<%=task.getTaskID()%>"> 
               <br>
               <%
-                PrivateCommentData pc = (PrivateCommentData) (privateCommentsMap.get(task.getTaskID()));
-                String message = (String) pc.getMessage().trim();
+                String ID = (String) "box-" + task.getTaskID();
               %>
-              <div id="box-<%=task.getTaskID()%>" onload="init('box-<%=task.getTaskID()%>', '<%=message%>')" type="text"></div>
-              <input type="hidden" name="message-<%=task.getTaskID()%>" value="document.getElementById('box-<%=task.getTaskID()%>').value(outputData);"</input>
+              <div id="box-<%=task.getTaskID()%>" onload="init('<%=ID%>')" type="text"></div>
```
jcespejo: should this be `task.getTaskID()`?

src/main/webapp/user-private-comments.jsp
```diff
@@ -56,11 +57,10 @@
...
+              <input type="hidden" name="message-<%=task.getTaskID()%>" value="document.getElementById('<%=ID%>').value(outputData);"</input>
```
jcespejo: Correct idea, but won't work since `document.getElementById` is javascript and javascript needs to be in `<script>` tags to run.  What you can do instead, when `save()` is called, you can save populate the input box

src/main/webapp/user-private-comments.jsp
```diff
@@ -56,11 +57,10 @@
...
               <button onclick="save()" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
```
jcespejo: `save()` will need to know which box to work on since you have multiple boxes on this screen.

src/main/webapp/user-private-comments.jsp
```diff
@@ -20,19 +20,19 @@
...
+        holderId : "box-" + privateCommentTaskID,
         autofocus: true,
-        data: privateCommentsDataJson[privateCommentID],
+        data: privateCommentsDataJson[privateCommentTaskID],
         tools: {
           header: Header,
         },
       });
-      console.log("here 2 !");
+      return editor;
```
talgsternberg, line 33: Have this return an editor (which I will pass to save()) so it knows which box to save.

src/main/webapp/user-private-comments.jsp
```diff
@@ -20,19 +20,19 @@
...
     }
-    function save() {
+    function save(editor) {
```
talgsternberg, line 35: Pass editor so it knows which editor is being saved as there is one for every task.

src/main/webapp/user-private-comments.jsp
```diff
@@ -56,12 +56,9 @@
             <form action="/user-profile" method="POST">
               <input type="hidden" name="taskID" value="<%=task.getTaskID()%>"> 
               <br>
-              <%
-                String ID = (String) "box-" + task.getTaskID();
-              %>
-              <div id="box-<%=task.getTaskID()%>" onload="init('<%=ID%>')" type="text"></div>
-              <input type="hidden" name="message-<%=task.getTaskID()%>" value="document.getElementById('<%=ID%>').value(outputData);"</input>
-              <button onclick="save()" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
+              <div id="box-<%=task.getTaskID()%>" onload="init(<%=task.getTaskID()%>)" type="text">Here</div>
```
talgsternberg, line 59: "Here" appears where editor should be in this div.

src/main/webapp/user-private-comments.jsp
```diff
@@ -56,12 +56,9 @@
...
+              <button onclick="save(init(<%=task.getTaskID()%>))" class="deep-button submit-button" type="submit" class="deep-button">Update Comment</button>
```
talgsternberg, line 60: Pass init(taskID) which returns editor to save().

src/main/webapp/user-private-comments.jsp
```diff
@@ -56,12 +56,9 @@
...
+              <input type="hidden" name="message-<%=task.getTaskID()%>" value="save()"</input>
```
talgsternberg, line 61: After save is called onclick of the button above, populate the input value with the value of save (which I think should be outputData?

src/main/webapp/user-private-comments.jsp
jcespejo: calling init() again here would probably reinitialize the box (would reset to the beginning state), so you'd loose any edits you've done in the data.

Alternative approach would be have init save the editor in a map, then save would lookup the editor from the map:

```
var editors = {}
function init(pcID) {
 // do stuff to init box
  editors[pcID] = editor;
}

function save(pcID) {
  editor = editors[pcID];
  // do save stuff to editor
}
```

src/main/webapp/user-private-comments.jsp
jcespejo: yep.  line [37 specifically](https://github.com/googleinterns/step171-2020/pull/94/files/badc7c173a80c0589f4702beac544d07cd686fd4#diff-243abfa42a44d81d97be6c5af0c05900R37), instead of just console.log there, that would be the correct place to do `document.getElementById('message' + pcID).value(outputData)`

PR #96, opened 2020-09-01
NotificationData class

godsfavours:
- Implemented PR suggestions, namely adding users by fake emails
- Created NotificationData class for notifications

### DIFF
--- a/src/main/java/com/rtb/projectmanagementtool/notification/NotificationData.java
+++ b/src/main/java/com/rtb/projectmanagementtool/notification/NotificationData.java
@@ -1,0 +1,160 @@
+package com.rtb.projectmanagementtool.notification;
+
+import com.google.appengine.api.datastore.Entity;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
+import java.util.HashSet;
+
+public class NotificationData {
+  private static final String PROPERTY_MESSAGE = "message";
+  private static final String PROPERTY_TIME_STAMP = "timeStamp";
+  private static final String PROPERTY_USERS = "users";
+
+  private long id;
+  private String message;
+  private Date timeStamp;
+  private HashSet<Long> users; // users to display this notification for
+
+  public NotificationData(String message, HashSet<Long> users) {
+    this.message = message;
+    this.users = users;
+    this.timeStamp = new Date();
+  }
+
+  public NotificationData(Entity entity) {
+    this.id = (Long) entity.getKey().getId();
+    this.message = (String) entity.getProperty(PROPERTY_MESSAGE);
+    this.timeStamp = (Date) entity.getProperty(PROPERTY_TIME_STAMP);
+
+    Collection entityProperty = (Collection) entity.getProperty(PROPERTY_USERS);
+    if (entityProperty != null) {
+      this.users = new HashSet<Long>((ArrayList<Long>) entityProperty);
+    } else {
+      this.users = new HashSet<Long>();
+    }
+  }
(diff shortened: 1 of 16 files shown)


src/main/java/com/rtb/projectmanagementtool/project/AddUserToProjectServlet.java
```diff
@@ -24,12 +25,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
     Long projectId = Long.parseLong(request.getParameter("project"));
-    String userInviteCode = request.getParameter("userInviteCode");
+    String userEmail = request.getParameter("userEmail");
     String userRole = request.getParameter("userRole");
 
     // create objects
     ProjectData project = projectController.getProjectById(projectId);
-    UserData user = userController.getUserByInviteCode(userInviteCode);
+
+    // need to call getEveryUser and iterate through the list to find user b/c email field is not
```
jcespejo, line 34: If you have made up emails, feel free to store than in the db.

src/main/java/com/rtb/projectmanagementtool/notification/NotificationData.java
```diff
@@ -0,0 +1,160 @@
...
+    Collection entityProperty = (Collection) entity.getProperty(PROPERTY_USERS);
+    if (entityProperty != null) {
+      this.users = new HashSet<Long>((ArrayList<Long>) entityProperty);
+    } else {
+      this.users = new HashSet<Long>();
+    }
+  }
+
+  /** @return the entity representation of this class */
+  public Entity toEntity() {
```
jcespejo, line 39: Will this be stored in the datastore?  I wonder if there's maybe another approach we can take.  "Notification" I feel by definition should be temporary.  It's a response to an event, and should no longer be available once delivered.  Like if you receive an email notification from gmail, the notification itself is probably not stored, but what's stored is the email and a flag specifying the email is unread.

godsfavours: I'm thinking the notifications will have persistent behavior, so they would need to be stored in datastore (even if its just for one user). Notifications can then be removed when a user manually clears it or the event that triggered the notification is overrided (ex. a user creates a task in a project -> create notification, that task is deleted -> remove notification if user hasn't already manually cleared it). I'm not sure how to approach this if its made temporary other than adding some real-time functionality

godsfavours: I added some stuff on design doc to explain my current implementation idea
I could automatically remove the notifications from datastore after some time so that the database doesn't get filled with a lot of notifications

jcespejo: > a user creates a task in a project -> create notification, that task is deleted -> remove notification if user hasn't already manually cleared it

I would count this as 2 separate events.  for example if the task was created, then someone else worked on the task, then the task was deleted, that person previously working on it would wonder what happened to the missing task.  notifications here would help since they would know that it was deleted.

>  so they would need to be stored in datastore

Storing temporarily in cache could be one other option.  if we need quicker reads/writes plus if it's temporary (expires after some time or after a user accumulated X number of events), cache could provide better performance than the datastore

godsfavours: Ah I see. I'll use datastore for the first pass and then maybe in my free time after the internship I'll see about implementing it with memcache

